PR #313, opened 2024-02-26
Combine transmission logic

cliffckerr:
Sorry to add to the glut of open PRs, but here is an experiment that combines the transmission logic of single and multi-RNG. I haven't tested it extensively, but I believe it is as common random number safe as `main` (i.e., is affected by https://github.com/amath-idm/starsim/issues/299, but can be fixed by the same fix as `main`). With this approach, multi-RNG is not appreciably slower than single-RNG.


cliffckerr:
Closed by #392

PR #339, opened 2024-02-29
Add simpleabm

cliffckerr:
Update: SimpleABM now lives at http://github.com/starsimhub/simpleabm. This PR instead provides a fix to `StaticNet`, so it works with no input arguments.

Implements a SimpleABM module, which is an as-simple-as-reasonable SIR (for now) ABM, including a spatial network.

Definitely not the PR I thought I'd be making now, but I ended up doing this for debugging Starsim's SIR model, and I feel like it might be useful for (a) debugging to know what the "simple" answer should be (eg do my SIR parameters make sense?), (b) rapid prototyping (eg a new network structure), (c) teaching (eg a rough approximation of what Starsim itself is doing).

Closes https://github.com/amath-idm/starsim/issues/124


RomeshA:
If it was in this repo, then putting it somewhere like the `tests` folder would make more sense, it is quite confusing for it to sit at the top alongside `starsim`. But I think a separate repo under starsimhub would be my preference too

menriquez-IDM:
Maybe having a `starsim-sandbox` or `starsim-samples` will be a good place to share implementations like this.  I see a lot of value on keeping them around in a visible spot.

PR #349, opened 2024-03-02
Extend RNG options to three, addressing issue #348

### LINKED ISSUE #348: Three options for rng
Currently, we have a boolean `multirng` option which defaults to True.

When True, distributions use the centralized numpy random number generator.

When False, a separate MultiRNG is used for each generator.

We need a third option wherein each distribution gets a _separate_ generator, but they're all just default numpy streams.

daniel-klein:
Renames and extends the `multirng` option in settings, now called 'rng'.

Set how random numbers are handled in Starsim with three options:
* "centralized" uses the centralized numpy random number generator for all distributions.
* "single" uses a separate (SingleRNG) random number generator for each distribution.
* "multi" uses a separate (MultiRNG) random number generator for each distribution.

"single" is the default value for this option.

In comparing two simulations, the "single" option may be slightly better than the "centralized" option without any real disadvantages. Only "multi" can achieve full common random number (CRN) coherence, but is more computationally expensive and only produces CRN results when using CRN-safe code.

In practice, devtest_rng.py shows that multi > single > centralized, but that single and centralized have nearly identical performance. So this PR switches the default to single, which uses the SingleRNG under the hood to track seeds for each stream.

Closes #348


cliffckerr (commented):
Let's merge this and then compare with `refactor-rngs`.

cliffckerr (approved):
Looks good, will merge for now

PR #369, opened 2024-03-08
Updates/networks

pausz:
Small PR:

+ StaticNet raises appropriate error if it receives [nx null_graph generator](https://networkx.org/documentation/stable/reference/generated/networkx.generators.classic.null_graph.html) -- this genrator does not take `n` as argument. 

+ NullNet is a specific instance of a network with a single contact per agent (self-connection) with weight zero. Main motivation to have this class is debugging purposes, models that do not use networks but may in the future (FPsim?); and perhaps enabling some network-based visualisations for models that don't really use networks. 

+ Empty networks (empty contact arrays) can be achieved with 
```python
empty_net_static = ss.StaticNet(nx.empty_graph)
empty_net_rand = ss.RandomNet(n_contacts=0)
```

### DIFF
--- a/starsim/network.py
+++ b/starsim/network.py
@@ -360,8 +360,8 @@ class SexualNetwork(Network):
 
 
 # %% Specific instances of networks
+__all__ += ['StaticNet', 'RandomNet', 'NullNet', 'MFNet', 'MSMNet', 'EmbeddingNet', 'MaternalNet']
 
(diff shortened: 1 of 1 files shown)


cliffckerr (commented):
Thanks @pausz , this will conflict with the `StaticNet` changes in #339 , but shouldn't be too hard to fix

pausz:
@kaa2102 Resolved the merge conflicts. Sorry for the delay!

Issue #295, opened 2024-02-22
Remove and/or skip and/or overwrite states


cliffckerr:
@jamiecohen ran into an issue where she wanted `rel_sus` to be a derived property from two other states. But `rel_sus` is added automatically when `Infection` is initialized. Need easy ways to (a) skip or overwrite default states, (b) remove states.

cliffckerr:
Closing in favor of #480

PR #383, opened 2024-03-14
Remove duplicate states from SIR

daniel-klein:


cliffckerr (commented):
This looks fine, but ... I can imagine it becoming a crazy bookkeeping task if each class adds a hodgepodge of states, and you have to look at the definitions of 10 different classes to figure out what the full list of states are! (Especially since you can't later overwrite them, see #295). E.g. here, how would I be expected to know that `ti_infected` is defined in `Infection` but `ti_dead` is defined in `SIR`?

PR #385, opened 2024-03-14
Remove unused `conception_probs` from Pregnancy

daniel-klein:
Seems straightforward unless someone is using `conception_probs` elsewhere?


menriquez-IDM:
Unless there is a reference on somebody's non-commited branch, this seems to be a safe change. I didn't find any reference.

PR #386, opened 2024-03-14
Update docs landing page

cliffckerr:
Old:
image

New:
image


README.rst
```diff
@@ -1,14 +1,13 @@
 Starsim
 =======
 
+**WARNING! Starsim is still in the early stages of development. It is being shared solely for transparency and to facilitate collaborative development. It is not ready to be used for real research or policy questions.**
```
JSchripsema-IDM, line 4: You can format this as a warning box and not just bold text by using the following syntax:

```
.. warning::
   Starsim is still in the early....
```

docs/index.rst
```diff
@@ -1,9 +1,40 @@
+.. title:: Starsim
+
+
+.. include:: ../README.rst
```
JSchripsema-IDM, line 4: You should probably update the README to replace the sentence saying documentation isn't available to include a link to docs.idmod.org/projects/starsim

JSchripsema-IDM (approved):
Looks good, but I noted a couple issues to be updated.

docs/index.rst
menriquez-IDM: there is currently a white IDM logo that is only visible if you are using the Dark theme:
**Ligth Mode:**
image

**Dark Mode:**
image

maybe you want to include a fix for this issue as well.

JSchripsema-IDM: Good catch! I can send the logo file if you need it.

README.rst
cliffckerr: That renders nicely in the docs:
image

But unfortunately GitHub doesn't parse that directive:
image

docs/index.rst
cliffckerr: Nice catch! Fixed

PR #389, opened 2024-03-15
Fixing bugs encountered while working on scenario modeling

daniel-klein:
* SIR duration of infection now accounts for dt
* Reworked sir_vaccine to modify rel_sus instead of moving agents from susceptible to recovered.
* n_years no longer necessarily an integer
* yearvec had an extra year
* Convert network keys to lowercase
* In Interventions, CampaignDelivery now is just discrete timepoints. Interpolation and annual_prob options removed. Fixed timepoints as `ss.true` functionality differs in Starsim than previous models, now using sc.findnearest.
* Also in Interventions, adding `coverage_dist` to BasicVaccination, which was missing.


starsim/diseases/sir.py
```diff
@@ -72,8 +72,8 @@ def set_prognoses(self, sim, uids, source_uids=None):
         will_die = p.p_death.rvs(uids)
         dead_uids = uids[will_die]
         rec_uids = uids[~will_die]
-        self.ti_dead[dead_uids] = sim.ti + dur_inf[will_die]
-        self.ti_recovered[rec_uids] = sim.ti + dur_inf[~will_die]
+        self.ti_dead[dead_uids] = sim.ti + dur_inf[will_die] / sim.dt # Consider rand round, but not CRN safe
+        self.ti_recovered[rec_uids] = sim.ti + dur_inf[~will_die] / sim.dt
```
cliffckerr, line 76: Should this be `//`?

starsim/sim.py
```diff
@@ -188,7 +188,7 @@ def init_time_vecs(self):
         """
         Construct vectors things that keep track of time
         """
-        self.yearvec = sc.inclusiverange(start=self.pars.start, stop=self.pars.end + 1 - self.pars.dt,
+        self.yearvec = sc.inclusiverange(start=self.pars.start, stop=self.pars.end - self.pars.dt,
```
cliffckerr, line 191: Much as I love Sciris, this has evolved into a very convoluted way to replicate the behavior of `np.arange(self.pars.start, self.pars.end, self.pars.dt)!`

cliffckerr: I think the previous version was correct, based on the comment "# Includes all the timepoints in the last year".

Issue #340, opened 2024-02-29
Ensure delta distributions are simply supported


cliffckerr:
Should have a simple API and be performant (eg not creating another distribution with zero width)

cliffckerr:
Closed by #392

Issue #334, opened 2024-02-28
Update distribution label


cliffckerr:
Currently `f'{self.name}_{self.label}_{key}'`, but this duplicates the name most of the time. I think safe to always just use `self.label`, should be equal to name or more unique than name.

cliffckerr:
Closed by #392

PR #322, opened 2024-02-27
Combine crn

cliffckerr:
Merge of #307 and #313 

NB: changed base to `main` since now more closely related to that than #307

PR #392, opened 2024-03-17
Refactor RNGs

### LINKED ISSUE #243: Explore "hash & cache" RNG
Potentially a way to generate random numbers without having to generate extra.
- [x] Replace `pandas` with more direct/faster approach
- [x] Combine single and multi RNG
- [x] Don't remove slots / consider factor of 5 difference
- [x] Simplify/standardize hashing step
- [x] Enable alternatives to SciPy (eg numpy RNG, scikit-learn)
- [x] Simplify `ScipyDistribution`
- [x] Simplify nonstandard distributions with ICDFs

See also https://github.com/amath-idm/starsim/issues/299

New checklist:
- [x] Reimplement `test_random.py`
- [x] Reimplement `test_rngcontainer.py`
- [x] Reimplement `test_distributions.py`
- [x] Get most basic flow working
- [x] Re-enable SciPy distributions
- [x] Define explicit distributions
- [x] Figure out Bernoulli distributions
- [x] Figure out delta distributions
- [x] Refactor code with new API
- [x] Get basic results to match
- [x] Get tests working
- [ ] Check performance

Even newer:
- [x] Figure out difference in indices in demo
- [ ] Figure out slots
- [x] Figure out what lognorm/lognorm_mean are doing
- [x] Figure out rv_discrete replacement
- [x] Figure out SciPyHistogram
- [ ] Figure out why sc.prettyobj breaks Dists
- [ ] Figure out why microsim breaks repr

cliffckerr:
Apologies for the somewhat large PR!

**Note**: This PR relies on a just-released version of Sciris, so please update (`pip install sciris --upgrade`, or `pip install -e .` for Starsim).

## New RNGs & distributions
- Replaces `ss.SingleRNG()`, `ss.MultiRNG()`, `ss.ScipyDistribution()`, and `ss.ScipyHistogram()` with a single `ss.Dist()` class. The `random.py` and `distributions.py` submodules have been removed, and `dists.py` has been added.
- The `ss.Dist` class uses `np.random.default_rng()` rather than `scipy.stats` by default, although a `scipy.stats` distribution can be supplied as an alternative. While comparable in most cases, in some instances Numpy is up to 4x faster (including, critically, for Bernoulli distributions).
- `ss.Dist` has various tools for understanding what it's doing, including `dist.show_state(), `dist.plot_hist()`, etc.
- Also removes `ss.options.multirng` (this branch is equivalent to it being always on).
- Removes duplicate logic for transmission (`make_new_cases()`), using an approach similar to the previous `make_new_cases_singlerng()` logic (though we could instead use the `make_new_cases_multirng()` logic; it should be similar in terms of performance and outcome, though slightly more complex code).
- Adds new custom distributions such as `ss.choice()` and `ss.delta()`.
- These distributions can be called directly, e.g. `dist = ss.weibull(a=2); dist(5)` will return 5 random variates from a Weibull distribution.
- Instead of being manually initialized based on the name, the `Sim` object is parsed and all distributions will be initialized with a unique identifier based on their place in the object (e.g. `sim.diseases.sir.pars.dur_inf`), which is used to set their unique seed. (Note: this relies on Sciris 3.1.5, which hasn't been released yet, so `sc_nested.py` has been copied here as a temporary fix.)

### Examples
```py
# 1. Create and use a simple distribution

# Old:
import starsim as ss
import scipy.stats as sps
dist = sps.norm(loc=4, scale=3)
rng = ss.MultiRNG('normal')
rng.initialize(container=None, slots=None)
dist.random_state = rng
d = ss.ScipyDistribution(dist)
print(d.rvs(5))

# New:
import starsim as ss
d = ss.Dist('normal', loc=4, scale=3)
print(d(5)) # dist() is an alias to dist.rvs()

# or as a one-liner:
print(ss.normal(4,3)(5))


# 2. Register distributions with the sim (pseudocode)

# Old:
sim = ss.Sim(...)
sim.rng_container = ss.RNGContainer()
sim.rng_container.initialize(seed)
for module in sim.modules:
  items = list(module.pars.items()) + list(module.__dict__.items()) # Isn't recursive
  for key,val in items:
    if isinstance(val, ss.ScipyDistribution):
      val.initialize(sim)

# New:
sim = ss.Sim(...)
sim.dists = ss.Dists(sim)
dists = sim.dists.find_dists() # Parse full Sim object, link to all ss.Dist objects regardless of depth
for dist in dists:
  dist.initialize(sim)

# 3. Show all distributions in the sim

# Old:
sim.rng_container._rngs # Shows RNG names, but have to manually find the corresponding ScipyDistributions

# New:
sim.dists.dists # Shows distributions as well as RNGs
```

## Other changes
- This PR also fixes bugs with lognormal parameters, and makes it explicit whether the parameters are for the *underlying* normal distribution (`ss.lognorm_u()`, the Numpy/SciPy default, equivalent to `ss.lognorm_mean()` previously) or the "overlying" lognormal distribution (`ss.lognorm_o()`, equivalent to `ss.lognorm()` previously).
- Renames `ss.dx`, `ss.tx`, `ss.vx` to `ss.Dx`, `ss.Tx`, `ss.Vx`.
- Removed `set_numba_seed()` as a duplicate of `set_seed()`.

## FAQ
- **Q: Why is this PR so big? Aren't PRs meant to be small?**
  - A: Yes, but distributions are central to how Starsim works, so it wasn't possible to make incremental changes. I promise I opened this PR as soon as I could get the tests to pass!
- **Q: Code is supposed to be modular, so why combine ``ss.MultiRNG`` and ``ss.ScipyDistribution`` into a single class?**
  - A: These objects were only ever used together, so creating them sepearately created extra housekeeping work.
- **Q: Without `ss.options.rng`, how can we compare results with and without CRN?**
  - A: Using different seeds for sims will give a pretty good approximation of what using a centralized RNG would look like. Alternatively, you could parse the `Dist` instances and replace all the RNGs with the same one (eek!). From a user perspective, keeping an option that we do not recommend using is confusing.
- **Q: How is performance on this branch?**
  - A: Not as fast as I'd like. Drawing random numbers is as fast or faster than before (up to 4x faster for Bernoulli distributions, for example), but overall performance is about the same as `multirng='multi'` was before. Going to spend some time figuring out why.
- **Q: What happened with slots?**
  - A: These still need to be reimplemented (currently, UIDs are being treated as slots, which is valid except in the case of different numbers of births between sims). However, first, I'd like to write a test that fails with the current implementation, and demonstrate that adding slots back in fixes it. (There is `test_repeat_slots`, but just showed that the same random numbers are drawn for each slot, whereas we want to show that simulations diverge without using slots but don't diverge when using them.)
- **Q: What else still needs to be done beyond this PR?**
  - A: See #395. Main issues are performance (including profiling), slots (including writing tests), simplifying how distributions are initialized, and various small "todo" comments for refactoring the code.

Closes #243, #316, #348, #312, #334, #340, #313, #322. Opens #395.


starsim/diseases/gonorrhea.py
```diff
@@ -22,14 +22,14 @@ def __init__(self, pars=None, par_dists=None, *args, **kwargs):
 
         # Parameters
         pars = ss.omergeleft(pars,
-            dur_inf_in_days = ss.lognorm(s=0.6, scale=10),  # median of 10 days (IQR 7–15 days) https://sti.bmj.com/content/96/8/556
+            dur_inf_in_days = ss.lognorm_o(mean=10, stdev=0.6),  # median of 10 days (IQR 7–15 days) https://sti.bmj.com/content/96/8/556
```
cliffckerr, line 25: Think this was a bug

daniel-klein:
Good, I like the simplifications this enables!

Just starting to form thoughts and questions:
* If a users wants to run a simulation with two of the same module (e.g. RSV-A and RSV-B), won't the trace be identical, and hence repeated random numbers? It seems like we'd need to use the name/label in addition to the trace.
* We need slots, agent uids alone doesn't achieve CRN. (I see how slots could be provided instead of uids, might just need to clean up the interface?)
* I suspect callable parameters are not working correctly.
 - It seems the call only occurs when the distribution is created (`make_dist`), so a user would need to regenerate the distribution on every timestep to use a callable with parameters that vary (e.g. age)
 - The key resolution of callable parameters with non-trivial distributions seems to have been lost (although I am still familiarizing myself with this PR, so apologies if I have simply missed it).
* Transmission
 - The approach in this PR calls `rng_source` and `rng_target` twice for each network, breaking CRN if the max slots differ at any point in time.
 - I believe this code will raise an exception if `strict` is enabled?
 - Also just thinking through what will happen when comparing simulations, one with network B and the other with networks A and B.
* Regarding performance, the unified approach to transmission uses a lot more random numbers that is needed if not attempting CRN, especially if slot_scale is large

cliffckerr:
Thanks @daniel-klein , very helpful comments! I agree that a lot of these are likely to be issues. I can work on them, but if you could help write tests for them (that fail on this branch) that would help me fix them. To respond to your specific comments:

### Won't the trace be identical for two modules with the same name?

No, because it's based on the location within the python object (via `sc.iterobj()`), and two objects can't have the same "physical" location. For example:
```py
import starsim as ss

sir1 = ss.SIR(name='sir1')
sir2 = ss.SIR(name='sir2')

sim = ss.Sim(pars=dict(networks='random'), diseases=[sir1, sir2])
sim.run()
sim.dists.dists.keys()

["('people', 'age_data_dist')",
 "('pars', 'sir1', 'dur_inf')",
 "('pars', 'sir1', 'dur_inf', 'module', 'rng_target')",
 "('pars', 'sir1', 'dur_inf', 'module', 'rng_source')",
 "('pars', 'sir1', 'init_prev')",
 "('pars', 'sir1', 'p_death')",
 "('pars', 'sir2', 'dur_inf')",
 "('pars', 'sir2', 'dur_inf', 'module', 'rng_target')",
 "('pars', 'sir2', 'dur_inf', 'module', 'rng_source')",
 "('pars', 'sir2', 'init_prev')",
 "('pars', 'sir2', 'p_death')"]
```

You can't create two diseases of the same type with the same name anyway (`ValueError: Cannot add object "sir" since already present in ndict "{self._type"} with keys: sir`), but if you do that, everything else is handled automatically. Even if you somehow could (e.g. they got appended to a list), the trace would still work.

### Callable parameters

- These are working insofar as `test_distributions.py:test_callable()` passes. But there might be more to do here. I believe that callable parameters works with arbitrary distributions: they're called, which creates an array, which is used as the parameter for the distribution for those distributions which support it. 
- `dist.make_dist()` is called every time `dist.rvs()` is called. Except in the case of a frozen SciPy distribution, this is not creating a new distribution, but just getting an attribute from `np.random.default_rng()`, so is fast. You're right though that this doesn't work with frozen SciPy distributions. My personal vote would be to not support the combination of callable parameters with frozen distributions -- I believe this is working against how frozen distributions are supposed to be used, which may be why the code required was so complex.

### Transmission

Yes, we could probably optimize how the random numbers are called (I did have a `dist.jump()` in the loop, but took it out -- we could restore it however). For slots, I think `max_slots = n_agents` might be fine? We can check what performance difference it makes. But my thinking is that with ~100k agents, and new agents assigned at random, the probability of two agents sharing the same slot and experiencing the same disease challenge on the same timestep is very slight. And even if that does happen, would it actually affect the results? We should write a test (maybe a devtest) to check.

kaa2102:
@cliffckerr I think you already addressed most if not all of @daniel-klein's comments. Regarding attainment of CRN, instead of `max_slots=n_agents`, I think we need slack in the number of slots. Could we allow `max_slots=random.rand(n_agents+1,1000) `based on the total number agents to be created in the simulation?

daniel-klein:
Thanks for the nice responses, @cliffckerr!

The module "trace" is really nice, yes!

But the callable parameters still will not work, I can explain when time permits and will create a test to demonstrate. (The code will run, but is technically very wrong.)

kaa2102 (approved):

daniel-klein:
So I can't fully create the test I want due to the lack of slots here, but it's something like this:

```python
def test_callable():
    def death_prob(context, sim, uids):
        p = np.ones(len(uids))
        p[-1] = 0 # Last agent has zero prob
        return p
    
    d = ss.bernoulli(p = death_prob)

    # HERE IS WHAT I WANT:
    uids = [0,1,2] # UIDs are always unique
    slots = [100,205,205] # Slots corresponding to the above UIDs can be repeated
    #draw = d.urvs(uids) # <- Argument is UIDs, slots should be used under the hood

    # BUT TO SHOW THE TEST...
    draw = d.urvs(slots) # <- Argument should be UIDs, but instead put slots here for the purpose of this test

    assert draw[1] != draw[2], 'Draws with different parameter values should be different, even if slot is shared'

    return draw
```

This test fails. The user should be expected to operate on the UID space, saying which agents are relevant. Behind the scenes, slots are used to do CRN stuff. Slots are not unique, but parameters _may_ differ between two agents that happen to have the same slot. Two agents with the same slot but different distribution parameters cannot receive the same result! This edge case is handled in main, but not here.

cliffckerr:
I see what you're saying ... I feel this is more a test of slots (and specifically repeated slots) rather than callable parameters though, right? If you replace `slots` with `uids` then the test passes, as it's similar to `test_distributions.py:test_callable()`.

But you're right that without slots, CRN isn't fully implemented, so this is definitely an important feature to add back in!

daniel-klein:
`make_dist` calls `lognorm_convert` for `lognorm_o` assuming the `mean` and `stdev` parameters are scalars, not callables (which are handled later). Would be nice to allow callables here!

starsim/dists.py
```diff
@@ -0,0 +1,519 @@
...
+        # Tidy up
+        self.called += 1
+        if self.strict:
+            self.ready = False
+        if uids is not None:
+            rvs = rvs[uids]
+            
+        return rvs
+    
+    def urvs(self, uids):
```
RomeshA, line 390: Maybe it could be clearer (it's not obvious to me, at least in briefly looking) why `urvs`is a different function given that `rvs` has a `uids` argument. It would be clearer if either `rvs` accepted only a scalar number of samples, or if `rvs` just treated the `size` argument differently depending on if it was a scalar or an array (where the latter could assume it was slots?)

tests/test_dists.py
```diff
@@ -0,0 +1,165 @@
...
+    sc.figlayout()
+    return fig
+
+
+# %% Define the tests
+def test_dist(m=m):
+    """ Test the Dist class """
+    sc.heading('Testing the basic Dist call')
+    dist = ss.Dist('random', 'test')
+    dist.initialize()
```
robynstuart, line 35: I am still not understanding why we need to initialize dists if we're not passing a sim/obj

tests/test_random.py
```diff
@@ -6,122 +6,106 @@
...
-
-
-def make_rng(slots=5, base_seed=1, name='Test', **kwargs):
-    rng_container = ss.RNGContainer()
-    rng_container.initialize(base_seed=base_seed)
-    rng = ss.RNG(name, **kwargs)
-    if ss.options.rng in ['single', 'multi']:
-        rng.initialize(rng_container, slots=slots)
-    return rng
+def make_dist(seed=1, name='test', **kwargs):
```
robynstuart, line 12: i'm not sure why we need this in addition to test_dists

tests/test_random.py
```diff
@@ -6,122 +6,106 @@
...
+def test_random(n=n):
     """ Simple random draw """
-    sc.heading('test_random: Testing simple random draw from a RNG object')
-    draws = rng.random(n)
+    sc.heading('Testing simple random draw from a RNG object')
+    
+    dist = make_dist()
+    draws = dist(n)
     print(f'Sampled {n} random draws', draws)
     assert len(draws) == n
```
robynstuart, line 27: again, this is repeated from test_dists, no? I realise why this matters is that often I look to the tests for guidance on the expected syntax. And with the new distributions implementation, I won't know whether to look in test_distributions, test_dists, test_random, or test_rngcontainer, all of which contain versions of the same test

starsim/dists.py
```diff
@@ -0,0 +1,519 @@
...
+        errormsg = f'Cannot create a lognorm_o distribution with mean≤0 (mean={mean}); did you mean to use lognorm_u instead?'
+        raise ValueError(errormsg)
+    std2 = stdev**2
+    mean2 = mean**2
+    under_std = np.sqrt(np.log(std2/mean2 + 1)) # Computes stdev for the underlying normal distribution
+    under_mean  = np.log(mean2 / np.sqrt(std2 + mean2)) # Computes the mean of the underlying normal distribution
+    return under_mean, under_std
+
+
+def find_dists(obj, verbose=False):
```
RomeshA, line 42: There are some edge cases that might be problematic involving circular references - I'd quite like to be able to have bidirectional references between modules but I think this doesn't get detected properly by `iterobj`, one fix could be to cache the objects it's previously visited?

starsim/sim.py
```diff
@@ -102,10 +92,9 @@ def initialize(self, reset=False, **kwargs):
         self.validate_dt()
         self.init_time_vecs()  # Initialize time vecs
         ss.set_seed(self.pars.rand_seed)  # Reset the random seed before the population is created
-        set_numba_seed(self.pars.rand_seed)
 
         # Initialize the core sim components
-        self.rng_container.initialize(self.pars.rand_seed + 2)  # +2 ensures that seeds from the above population initialization and the +1-offset below are not reused within the rng_container
+        self.dists.initialize(obj=self, base_seed=self.pars.rand_seed + 2)  # +2 ensures that seeds from the above population initialization and the +1-offset below are not reused within the rngs
```
robynstuart, line 97: idk if I'm being paranoid, but should we control the +2 some other way... would this break if someone came along and added another seed? Should we have some kind of counter we increment?

starsim/sim.py
```diff
@@ -117,6 +106,7 @@ def initialize(self, reset=False, **kwargs):
         self.init_analyzers()
 
         # Perform post-initialization validation
+        self.dists.initialize(obj=self, base_seed=self.pars.rand_seed + 2)  # TEMP # TODO Should not be here!!!
```
robynstuart, line 109: yes I am confused why this is here :) it's on line 52, 97, and again here.

starsim/dists.py
```diff
@@ -0,0 +1,520 @@
...
+            self.base_seed = base_seed
+        sim = sim if (sim is not None) else self.sim
+        obj = obj if (obj is not None) else self.obj
+        if obj is None:
+            errormsg = 'Must supply a container that contains one or more Dist objects'
+            raise ValueError(errormsg)
+        self.dists = find_dists(obj)
+        for trace,dist in self.dists.items():
+            if not dist.initialized or force:
+                dist.initialize(trace=trace, seed=base_seed, sim=sim)
```
robynstuart, line 84: Are we somehow ensuring unique names during this process?

starsim/dists.py
```diff
@@ -0,0 +1,520 @@
...
+        self.check_seeds()
```
robynstuart, line 85: Confused again :) in the line above we seem to be setting the seed for each dist to be the same, but then we check that they are all different... ?

starsim/dists.py
```diff
@@ -0,0 +1,519 @@
...
+            dist.rvs(10) # Return 10 normally distributed random numbers
+        """
+        self.dist = dist
+        self.name = name
+        self.kwds = sc.dictobj(kwargs)
+        self.seed = seed # Usually determined once added to the container
+        self.offset = offset
+        self.module = module
+        self.sim = sim
+        self.strict = strict
```
RomeshA, line 169: One thing I'm not sure about yet here is the value vs risk of having strictness set at the distribution level. Is there a risk that a user who wants to be completely RNG safe finds it hard to be able to guarantee that they're not accidentally doing something unsafe? The issue being that the dists default to everything being unsafe and everything working, so it could be quite easy for everything to look like it's running nicely even though it's unsafe. On the other hand, I can see the appeal for testing/development purposes. Maybe the main thing would be having a single point (e.g., inside `Dists.inititialize()` for example) where all of the distributions in a `Sim` are checked for strictness if that was desired?

starsim/dists.py
```diff
@@ -0,0 +1,519 @@
...
+        # Draw using UIDs rather than a fixed size
+        uids = np.array([1,2,4,9])
+        ss.Dist('bernoulli', p=0.5).urvs(uids)
+    """
+    def __init__(self, dist=None, name=None, seed=None, offset=None, strict=False, module=None, sim=None, **kwargs): # TODO: switch back to strict=True
+        """
+        Create a random number generator
+        
+        Args:
+            dist (str): the name of the (default) distribution to draw random numbers from, or a SciPy distribution
```
RomeshA, line 148: Can we give an example somewhere of how to create an entirely custom distribution? Does this need to be done as a class inheriting from Scipy's distribution class?

tests/test_distributions.py
```diff
@@ -1,189 +1,148 @@
...
+    low  = np.array([1, 100]) # Low
+    high = np.array([3, 125]) # High
 
-    dist = sps.uniform(loc=loc, scale=scale)
-    dist.random_state = rng
-
-    d = ScipyDistribution(dist)
-    draws = d.rvs(uids)
+    d = ss.uniform(low=low, high=high).initialize()
+    draws = d.urvs(uids)
```
robynstuart, line 98: again pretty minor comment, but `d.rvs(3)` could return a more helpful error.

starsim/dists.py
```diff
@@ -0,0 +1,519 @@
...
+        return
+    
+    def set(self, dist=None, **kwargs):
+        """ Set (change) the distribution type, or one or more parameters of the distribution """
+        if dist:
+            self.dist = dist
+        self.kwds = sc.mergedicts(self.kwds, kwargs)
+        return
+    
+    def make_dist(self, size, uids=None):
```
RomeshA, line 295: Am I understanding correctly that this is called every time a sample is drawn? Is this function cheap enough for that? Otherwise could this be done once at initialization?

starsim/dists.py
```diff
@@ -0,0 +1,519 @@
...
+        """ Ensure the supplied dist and parameters are valid, and initialize them; called automatically """
+        dist = self.dist # The name of the distribution (a string, usually)
+        kwds = self.kwds.copy() # The actual keywords; shallow copy, modified below for special cases
+        
+        # Main use case: handle strings, including special cases
+        if isinstance(dist, str):
+            
+            # Handle lognormal distributions
+            if dist == 'lognorm_o': # Convert parameters for a lognormal
```
RomeshA, line 304: It feels a bit odd to hardcode this level of functionality here, is there a more canonical way this could be implemented? e.g., if we're recommending people inherit from scipy's distribution class, implementing our versions of that lognormal distribution, and then using those when calling `ss.lognorm_o`?

I guess the same would be true of the other branches (for bernoulli and delta) if that was something that could be handled by having the appropriate distribution constructed beforehand

starsim/dists.py
```diff
@@ -0,0 +1,520 @@
...
+            key = str(trace)
+            out[key] = val
+            if verbose: print(f'  {key} is a dist ({len(out)})')
+    return out
+
+
+class Dists:
+    """ Class for managing a collection of Dist objects """
+
+    def __init__(self, obj=None, base_seed=None, sim=None):
```
robynstuart, line 58: consider accepting args as more distributions? i thought this would work 
```
dd = ss.Dists(ss.uniform(), ss.lognorm_o()) # Doesn't complain
len(dd.dists). # Returns 1
```

starsim/disease.py
```diff
@@ -231,176 +229,83 @@ def init_results(self, sim):
...
-
-                new_cases_bool = np.random.random(
-                    len(a)) < p_transmit  # As this class is not common-random-number safe anyway, calling np.random is perfectly fine!
-                new_cases.append(b[new_cases_bool])
-                sources.append(a[new_cases_bool])
+                p_transmit = rel_trans[src] * rel_sus[trg] * beta_per_dt
+
+                slots_s = people.slot[src] # Slots for the possible source
+                slots_t = people.slot[trg] # Slots for the possible target
+                rvs_s = self.rng_source.rvs(size=np.max(slots_s) + 1)[slots_s]
```
RomeshA, line 290: I'm probably missing something obvious, but I had thought this would be able to just call `urvs` directly?

tests/test_dists.py
```diff
@@ -0,0 +1,165 @@
...
+
+    # Create the objects twice
+    for i in range(2):
+        
+        # Create a complex object containing various distributions
+        obj = sc.prettyobj()
+        obj.a = sc.objdict()
+        obj.a.mylist = [ss.random(), ss.Dist('uniform', low=2, high=3)]
+        obj.b = dict(d3=ss.weibull(a=2), d4=ss.delta(v=0.3))
+        dists = ss.Dists(obj)
```
robynstuart, line 82: trying to understand the automatic naming here:
```
# With one dist, it gets called 'root' (why?)
dd = ss.Dists(ss.uniform())
[d.name for d in dd.dists.values()]
Out[79]: ['root']

# With 2 dists, they get names that are stings of tuples of ints...  
dd = ss.Dists([ss.uniform(), ss.lognorm_o()])
[d.name for d in dd.dists.values()]
Out[81]: ['(0,)', '(1,)']
```

is this how it's supposed to be?

starsim/dists.py
RomeshA: Ah I see you've already discussed that previously around handling callable parameters and this generally being cheap. I do still wonder about the cost of the other branches (e.g., it seems unnecessary to be creating new lambda functions with each call) but that's not obviously a problem depending on use patterns

robynstuart (commented):
Nice! I've taken a first pass - may have more comments later as I keep playing around with it. Most of my comments are relatively minor, but my biggest concern is resolving @daniel-klein's comment about callable parameters. Also, there's a lot of repetition in the tests and I feel like we're missing a few core functionality tests. Will keep experimenting though.

RomeshA:
These changes streamline things a lot, I like it! Agree with @daniel-klein that would be great to see how it looks with slots in there though, since implementing that cleanly is a big part of the challenge

daniel-klein:
Distributions created as part of pars that eventually feed into `convert_plugins` get an incorrect trace, and therefore a different seed compared to if the same distribution was supplied to the module directly. I propose that we remove `convert_plugins` and the dict-based parameterization rather than changing how the trace is handed, it gets quite complicated to initialize otherwise.

cliffckerr:
> Distributions created as part of pars that eventually feed into `convert_plugins` get an incorrect trace, and therefore a different seed compared to if the same distribution was supplied to the module directly. I propose that we remove `convert_plugins` and the dict-based parameterization rather than changing how the trace is handed, it gets quite complicated to initialize otherwise.

Agree, we should ensure this is fixed as part of fixing #284, #298, and #301

starsim/dists.py
cliffckerr: It does via a `deepcopy`-like `memo`, though this functionality was just added in Sciris 3.1.5:
https://github.com/sciris/sciris/blob/main/sciris/sc_nested.py#L391

starsim/dists.py
cliffckerr: Tracking in #408

starsim/dists.py
cliffckerr: Tracking in #408

starsim/dists.py
cliffckerr: Tracking in #408

starsim/dists.py
cliffckerr: It does seem fine since the distributions are only called a few times per timestep, but I am planning on doing more benchmarking to be sure.

starsim/dists.py
cliffckerr: Tracking in #408. I agree it's not perfect, but I think it's an improvement on the previous version which _looked_ like it was using SciPy's `lognorm`, but was actually transforming the parameters under the hood! I tried to make the difference clear that there are `lognorm_o` and `lognorm_u` functions, so one wouldn't mistakenly just use "lognorm" as the default (with different people making different assumptions about what that was), but not sure this choice  quite nails the explicitness-simplicity tradeoff, so open to suggestions!

starsim/disease.py
cliffckerr: Yes, this is legacy code. Tracking in #408

docs/api/index.rst
```diff
@@ -15,14 +15,13 @@ This page lists Starsim's full API, including all functions and classes.
    starsim.connectors
    starsim.demographics
    starsim.disease
-   starsim.distributions
+   starsim.dists
```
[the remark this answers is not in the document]
cliffckerr: I called it dists because I'd thought some people might want to use it directly, e.g. `ss.dists.random()`, but ... that's probably a silly reason. I'll rename it. Tracking in #408

starsim/demographics.py
```diff
@@ -9,6 +9,7 @@
 
 __all__ = ['BaseDemographics', 'Births', 'Deaths', 'Pregnancy']
 
+
 class BaseDemographics(ss.Module):
```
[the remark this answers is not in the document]
cliffckerr: Tracking in #408

starsim/disease.py
```diff
@@ -231,176 +229,83 @@ def init_results(self, sim):
...
+        Common-random-number-safe transmission code works by mapping edges onto
+        slots.
+        """
         new_cases = []
         sources = []
         people = sim.people
         betamap = self._check_betas(sim)
 
-        for nkey, net in sim.networks.items():
+        for i,nkey,net in sim.networks.enumitems():
```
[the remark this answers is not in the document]
cliffckerr: It had been used to reset the seed but I took this out. It needs to be refactored one way or another though! Tracking in #408

starsim/diseases/gonorrhea.py
```diff
@@ -87,7 +87,7 @@ def set_prognoses(self, sim, target_uids, source_uids=None):
 
         # Set natural clearance
         clear_uids = self.pars.p_clear.filter(target_uids)
-        dur = sim.ti + self.pars['dur_inf_in_days'].rvs(clear_uids)/365/sim.pars.dt # Convert from days to years and then adjust for dt
+        dur = sim.ti + self.pars['dur_inf_in_days'].urvs(clear_uids)/365/sim.pars.dt # Convert from days to years and then adjust for dt
```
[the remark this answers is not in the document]
cliffckerr: Tracking in #408

starsim/diseases/syphilis.py
```diff
@@ -264,7 +263,7 @@ def set_congenital(self, sim, target_uids, source_uids=None):
 
                 # Birth outcomes must be modified to add probability of susceptible birth
                 birth_outcomes = self.pars.birth_outcomes[state]
-                assigned_outcomes = birth_outcomes.rvs(uids)-uids  # WHY??
+                assigned_outcomes = birth_outcomes.urvs(uids)-uids  # WHY??
```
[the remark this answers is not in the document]
cliffckerr: Tracking in #408

starsim/dists.py
```diff
@@ -0,0 +1,520 @@
...
+
+def str2int(string, modulo=1e8):
+    """
+    Convert a string to an int via hashing
+    
+    Cannot use Python's built-in hash() since it's randomized for strings. While
+    hashlib is slower, this function is only used at initialization, so makes no
+    appreciable difference to runtime.
+    """
+    return int(hashlib.sha256(string.encode('utf-8')).hexdigest(), 16) % int(modulo)
```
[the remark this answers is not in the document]
cliffckerr: Fixed, now using `int.from_bytes()`

starsim/dists.py
```diff
@@ -0,0 +1,520 @@
...
+        uids = np.array([1,2,4,9])
+        ss.Dist('bernoulli', p=0.5).urvs(uids)
+    """
+    def __init__(self, dist=None, name=None, seed=None, offset=None, strict=False, module=None, sim=None, **kwargs): # TODO: switch back to strict=True
+        """
+        Create a random number generator
+        
+        Args:
+            dist (str): the name of the (default) distribution to draw random numbers from, or a SciPy distribution
+            name (str): the unique name of this distribution, e.g. "coin_flip" (in practice, usually generated automatically)
```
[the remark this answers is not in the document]
cliffckerr: Correct, they don't need to be unique. Tracking in #408

starsim/dists.py
```diff
@@ -0,0 +1,520 @@
...
+            seed (int): if supplied, the seed to use for this distribution
+            offset (int): the seed offset; will be automatically assigned (based on hashing the name) if None
+            module (ss.Module): if provided, used when supplying lambda-function arguments as parameters
```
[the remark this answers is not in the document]
cliffckerr: Will revise. Tracking in #408

starsim/dists.py
```diff
@@ -0,0 +1,520 @@
...
+            sim (ss.Sim): if provided, used when supplying lambda-function arguments as parameters
+            strict (bool): if True, require initialization and invalidate after each call to rvs()
```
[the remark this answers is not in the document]
cliffckerr: To catch errors with using an improperly initialized RNG. Tracking in #408

tests/test_distributions.py
```diff
@@ -1,189 +1,148 @@
...
-        rng = ss.MultiRNG('Uniform')
-        rng.initialize(container=None, slots=n)
-        dist.random_state = rng
-    d = ScipyDistribution(dist)
 
+def test_basic():
+    """ Basic scipy.stats test """
+    sc.heading('Test basic scipy.stats usage')
+    spsdist = sps.norm(loc=1, scale=1) # Make a distribution
+    d = ss.Dist(dist=spsdist).initialize() # Convert it to Starsim
```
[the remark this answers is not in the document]
cliffckerr: Correct. Tracking in #408

tests/test_distributions.py
```diff
@@ -1,189 +1,148 @@
...
-    rng = ss.MultiRNG('Uniform')
-    rng.initialize(container=None, slots=n)
-    dist = sps.uniform(loc=1, scale=4)
-    dist.random_state = rng
-    d = ScipyDistribution(dist)
-
-    uids = np.array([1,3])
-    draws = d.rvs(uids)
+    # Draw some samples
+    d.reset()
```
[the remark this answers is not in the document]
cliffckerr: Will add! Tracking in #408

tests/test_distributions.py
```diff
@@ -1,189 +1,148 @@
...
-        d.rng.initialize(container=None, slots=n) # Only really needed for testing as initializing the distribution will do something similar.
-
-    uids = np.array([1,3])
-    draws = d.rvs(uids)
-    print(f'Uniform sample for uids {uids} returned {draws}')
-
-    assert len(draws) == len(uids)
-    return draws
-
+    assert len(draws) == len(uids), 'Incorrect number of draws'
```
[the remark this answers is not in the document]
cliffckerr: Will tidy. Tracking in #408

tests/test_distributions.py
```diff
@@ -1,189 +1,148 @@
...
+    sim = sc.prettyobj()
+    sim.n = 10
+    sim.uid = np.arange(sim.n)
+    sim.age = np.random.uniform(0, 90, size=sim.n)
 
-    loc = lambda self, sim, uids: sim.people.age[uids] # Low
-    scale = 1 # Width, could also be a lambda
-    dist = sps.uniform(loc=loc, scale=scale)
+    # Define a parameter as a lambda function
+    loc = lambda module, sim, uids: sim.age[uids]
```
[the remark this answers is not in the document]
cliffckerr: Agree. Tracking in #408

tests/test_dists.py
```diff
@@ -0,0 +1,165 @@
+"""
+Test Dist and Dists
```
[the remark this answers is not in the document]
cliffckerr: Agree. Tracking in #408

tests/test_dists.py
cliffckerr: You're right, we don't (it was a holdover from strict=True). Tracking in #408

tests/test_random.py
cliffckerr: Tracking in #408

tests/test_random.py
cliffckerr: Lol brutally true. Tracking in #408

starsim/sim.py
cliffckerr: Tracking in #408

starsim/sim.py
cliffckerr: Tracking in #408

starsim/dists.py
cliffckerr: Yes, based on the distribution's location within the object (which in turn is determined by `sc.iterobj()`)

starsim/dists.py
cliffckerr: We are setting the same "user seed" for each, but this is then combined with the (unique) "offset seed" to create the actual seed, which should be unique.

RomeshA:
One thought (maybe just something we should be aware of) - does using the trace to assign RNG seeds make it hard to validate refactoring because changing the data structure could shift the seeds around and make it more common to be unable to validate that results are exactly equal?

tests/test_dists.py
cliffckerr: It is yes -- this is how `sc.iterobj()` parses the object and turns it into a series of traces. It probably makes more sense if you look at the an sim example:
```py
sim.dists.dists.keys()

 "('pars', 'syphilis', 'dur_primary')",
 "('pars', 'syphilis', 'dur_secondary')",
 "('pars', 'syphilis', 'dur_latent_temp')",
 "('demographics', 'pregnancy', 'pars', 'fertility_rate')",
 "('demographics', 'deaths', 'pars', 'death_rate')",
 "('networks', 'mfnet', 'pars', 'duration')",
 "('networks', 'mfnet', 'pars', 'participation')",
```

cliffckerr:
> One thought (maybe just something we should be aware of) - does using the trace to assign RNG seeds make it hard to validate refactoring because changing the data structure could shift the seeds around and make it more common to be unable to validate that results are exactly equal?

I don't think so, because it's based on dicts, so everything should be key-based rather than position-based. It's true that if you give the same modules different names, you'll get different results. But that's true of the implementation on `main` as well.

cliffckerr:
Remaining issues consolidated in https://github.com/starsimhub/starsim/issues/426

PR #401, opened 2024-03-20
Refactor RNG refactor

cliffckerr:


daniel-klein:
`urvs` takes uids --> slots and calls `rvs`, which later calls `make_dist` and ultimately callable like `make_death_prob_fn`, but by this point we have only slots, and not UIDS so stuff breaks.

Issue #402, opened 2024-03-20
Many disease modules are neglecting dt


daniel-klein:
Often see 
```python
ti_event = sim.ti + draw_from_duration_distribution
```
But this is mixing time time (ti) with time (the distribution draw).

daniel-klein:
Noticed yesterday that the measles model has parameters in days (according to the comment), but they are used as years. Addressed in #411, but that is a draft PR.

cliffckerr:
Merged

PR #403, opened 2024-03-20
Adding dt to disese modules

daniel-klein:
Adding dt to disese modules. Note, did not sc.randround to an exact ti, but rather left ti as a float computed as sim.ti + duration/sim.dt. This is in part because randround is not crn safe. The check is if event_ti < sim.ti, so by not rounding, events will be shifted to the next timestep. Hopefully dt is small enough relative to duration that this doesn't really matter.

### DIFF
--- a/starsim/diseases/cholera.py
+++ b/starsim/diseases/cholera.py
@@ -147,7 +147,7 @@ class Cholera(ss.Infection):
         p = self.pars
 
         # Determine when exposed become infected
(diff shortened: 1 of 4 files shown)


daniel-klein:
Addresses #402

cliffckerr (commented):
Would it be safer to use time rather than timestep here? Feels like a subtle bug and easy to make mistake to forget `sim.dt`. Plus, `ti_infected` is an int array, so will be immediately (and implicitly) cast to an int. Not sure how big the performance hit to go int → float → int is, probably pretty small?

PR #411, opened 2024-03-22
For comment: API simplification

daniel-klein:
* Remove par_dists. Bernoulli parameters are hard coded so users can simply enter floats that are interpreted as the 'p' value. Other distributions are now explicit, with new functionality for setting parameter values.
* Remove sim configuration by dict and corresponding `convert_plugins` logic. Users can still configure a Sim by dictionary simply by **pars.
* Moved `init_prev` and corresponding `set_initial_states` to `Disease`
* Distribution seed fixing w/r/t initialization
* General cleanup

### DIFF
--- a/docs/tutorials/tut_buildsim.ipynb
+++ b/docs/tutorials/tut_buildsim.ipynb
@@ -91,7 +91,7 @@
     "Now, let's look at a few useful ways to improve our model by extending these three components (people, networks, and diseases).\n",
     "\n",
     "## Making changes to our components\n",
-    "One of the main advantages of agent-based models is they allow you to capture heterogeneity between people. In real life, it's not realistic that everyone in a population has the same number of contacts with other people. Let's make our contact network more realistic by adding some variation here. For this, we'll use a Poisson distribution. The two lines below both do the same thing:"
+    "One of the main advantages of agent-based models is they allow you to capture heterogeneity between people. In real life, it's not realistic that everyone in a population has the same number of contacts with other people. Let's make our contact network more realistic by adding some variation here. For this, we'll use a Poisson distribution."
    ]
   },
   {
@@ -102,8 +102,7 @@
    },
    "outputs": [],
    "source": [
-    "network = ss.RandomNet(pars={'n_contacts': ss.poisson(4)})\n",
-    "network = ss.RandomNet(pars={'n_contacts': 4}, par_dists={'n_contacts': ss.poisson})"
+    "network = ss.RandomNet(pars={'n_contacts': ss.poisson(4)})"
    ]
   },
   {
--- a/starsim/demographics.py
+++ b/starsim/demographics.py
@@ -147,10 +147,6 @@ class Deaths(BaseDemographics):
             units = 1e-3,  # assumes death rates are per 1000. If using percentages, switch this to 1
         )
 
-        self.par_dists = ss.omergeleft(par_dists,
-            death_rate = ss.bernoulli
-        )
-
         # Process metadata. Defaults here are the labels used by UN data
         self.metadata = ss.omergeleft(metadata,
             data_cols = dict(year='Time', sex='Sex', age='AgeGrpStart', value='mx'),
@@ -160,7 +156,7 @@ class Deaths(BaseDemographics):
         # Process data, which may be provided as a number, dict, dataframe, or series
         # If it's a number it's left as-is; otherwise it's converted to a dataframe
         self.death_rate_data = self.standardize_death_data()
-        self.pars.death_rate = self.make_death_prob_fn
+        self.pars.death_rate = ss.bernoulli(self.make_death_prob_fn)
 
         return
 
@@ -267,11 +263,8 @@ class Pregnancy(BaseDemographics):
             units = 1e-3,          # Assumes fertility rates are per 1000. If using percentages, switch this to 1
         )
 
-        self.par_dists = ss.omergeleft(par_dists,
-            fertility_rate = ss.bernoulli,
-            maternal_death_rate = ss.bernoulli,
-            sex_ratio = ss.bernoulli
-        )
+        self.pars.maternal_death_rate = ss.bernoulli(self.pars.maternal_death_rate)
+        self.pars.sex_ratio = ss.bernoulli(self.pars.sex_ratio)
 
         # Process metadata. Defaults here are the labels used by UN data
         self.metadata = ss.omergeleft(metadata,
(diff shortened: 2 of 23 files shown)


daniel-klein:
Keen for discussion.

I much prefer the explicit distribution references, especially now that distribution parameters can be more easily modified, e.g. `sir.pars.dur_inf['scale'] = 6`

I also find this approach much less confusing to the users because there aren't multiple ways to configure a simulation. I do not understand the use case for `diseases="sir"` beyond something demo(), and even then, I'd like to teach users more extensible patterns. And we will eventually have more disease modules, and it might not be clear which "hiv" module will actually be used.

Please share thoughts, I know there are other opinions and I might be missing some important considerations.

cliffckerr (commented):
Thanks! Will definitely want to keep the fixes and tidying.

Re: the API changes, a couple comments:
- I agree we need to revamp how pars/kwargs are ingested and they need to be treated equivalently.  `convert_plugins` ended up being quite complicated, which is undesirable.
- However, I much prefer the simplicity and brevity of `ss.Sim(diseases='sir', networks='embedding')` compared to something more verbose that's functionally equivalent. I personally use this pattern all the time! I don't see a benefit of removing the simplest way of accessing basic functionality. Often I'll use `diseases='sir'` while I'm exploring different network properties, or `networks='random'` when diving deeper into a particular disease. I trust our users to understand our extensible patterns without those being forced upon them :)
- I also agree that having `pars` and `par_dists` isn't ideal. But nor do I like enforcing a mix of instantiated objects with simple scalar parameter settings like `dur_inf = ss.lognorm_o(mean=5, stdev=2)`. I think it will make things like calibration etc. harder to do.
- My personal preference would be to have something like `dur_inf = {'lognorm_o', 5, 2}` or `dur_inf = dict(type='lognorm_o', mean=5, stdev=2)`. This makes it easy to update, import, and export. It also means the sim doesn't have to carry around a half-initialized `Dist` object, waiting to get the sim and be properly initialized at a later point. (It should still _work_ to supply a full dist, but I personally think we should discourage this particular pattern.)

daniel-klein:
Thanks @cliffckerr, happy to keep iterating here.

* I would think the difference between `diseases='sir'` and `diseases=ss.SIR()` is not too much to ask of our users, and similar for networks. The class-based approach is the starting point for more advanced modeling, as it's straightforward to provide module-level parameters, whereas the string-based approach seems like a dead end. Considering it is just a few more characters to type to get into an extensible pattern, I removed support for the string-based approach in this PR.

* Regarding distributions, I love the readability provided by `dur_inf = ss.lognormal(mean=5, stdev=2)` in module code. It's clear to users that the default duration of infection is log-normally distributed with the specified parameters (_u and _o aside). Users can change parameters through the interface introduced in this PR, `dur_inf['mean'] = 3` or `dur_inf['stdev'] = 2`. It's clear what 3 and 2 are doing here, as opposed to a syntax like `dur_inf=5`, where the 5 might be doing something the user does not expect like setting the `stdev` of the underlying distribution rather than changing the `mean` or even setting a fixed value (point mass) of 5.

* I see you argument for `dict(type='lognorm_o', mean=5, stdev=2)`, but am not sure that encoding the information in a `dict` class is any better than encoding that same information in a `Dist` class. Both can be updated, imported, and exported - where the latter may have an advantage in parameter sanity checking and that the full state (including random state) can be saved and loaded. Strongly agree on revisiting initialization to ensure we get that right.
 
Overall, I think we should guide users to clear and explicit usage pathways via syntax, even if it ends up being a few characters longer, rather than providing shortcuts that could, in my opinion, lead to confusion or even model misconfiguration.

Wondering if other developers might have an opinion as well? cc @RomeshA @pausz @robynstuart @kaa2102

cliffckerr:
I agree that there's not a huge difference between `diseases='sir'` and `diseases=ss.SIR()`. But here's an example of where I see a big difference:
```py
pars = dict(diseases=['hiv', 'syphilis'], networks = ['mf', 'msm'])
s1 = ss.Sim(pars)
s2 = ss.Sim(pars, interventions=ss.Antenatal())
```
vs
```py
s1 = ss.Sim(diseases=[ss.HIV(), ss.Syphilis()], networks=[ss.MFNet(), ss.MSMNet()])
s2 = ss.Sim(diseases=[ss.HIV(), ss.Syphilis()], networks=[ss.MFNet(), ss.MSMNet()], interventions=ss.Antenatal())
# or
inputs1 = dict(diseases=[ss.HIV(), ss.Syphilis()], networks=[ss.MFNet(), ss.MSMNet()])
inputs2 = sc.dcp(inputs1)
s1 = ss.Sim(**inputs1)
s2 = ss.Sim(**inputs2, interventions=ss.Antenatal())
```
Passing around instantiated objects, when what is really desired is simply a configuration setting, is clunky both for the user (more typing) and the kernel (even in this simple example deepcopying ~100 KB code, only 16 bytes of which matter).

I obviously don't think we should _remove_ the ability to call `ss.HIV()` or `ss.lognorm(mean=5, stdev=2)` (though that's a funny example, speaking of potential for "confusion and model misconfiguration" -- is that the mean of the lognormal distribution, or the underlying normal distribution?!). I agree it's readable and I'm happy for users to use that if they want to! But I see little downside in supporting a simpler API when the user wants to do something simple, especially when it increases the likelihood of bugs, such as:
```py
diseases = 'sir'
s1 = ss.Sim(diseases=diseases)
s2 = ss.Sim(diseases=diseases) # works, not a bug

diseases = ss.SIR()
s1 = ss.Sim(diseases=diseases)
s2 = ss.Sim(diseases=diseases) # will not work -- can't reuse an instantiated object
```

daniel-klein:
Very good point about uses possibly reusing objects as a possible problem!

If this was a real modeling application of Starsim, I would likely structure the code like this:
```python
def make_sim(ANC_cov=None):
    # First create the diseases
    hiv = ss.HIV(pars=dict(...)) # HIV with some modified parameters
    syph = ss.Syphilis() # Syphilis with default parameters
    diseases = [hiv, syph]

    # Now create the networks
    mf = ss.MFNet(pars=dict(...)) # Heterosexual network
    msm = ss.MSMNet() # MSM network with default parameters
    nets = [mf, msm]

    intv = []
    if ANC_cov is not None and ANC_cov > 0:
        # Add an ANC intervention with the user-supplied coverage
        anc = ss.ANC(pars=dict(coverage=ANC_cov))
        intv.append(anc)
        label = f'ANC coverage {ANC_cov}' # Label accordingly
    else:
        label = 'Baseline'

    # Finally create and return the simulation instance
    sim = ss.Sim(diseases=diseases, networks=nets, interventions=intv, label=label)
    return sim

s1 = make_sim(ANC=0) # Baseline without ANC
s2 = make_sim(ANC=0.8) # 80% ANC coverage
```

This approach, while _much_ more verbose, has lots of hooks for user parameters, modifications, and customization. It can easily feed into `sc.parallelize` or `MultiSim` as well. My expectation is that most Starsim applications to look something like this. I fear that users will develop bad habits if we provide too many shortcuts.

cliffckerr:
I totally agree in the case of a "real modeling application". But personally -- and I'm pretty sure I'm not the only one who works like this! -- even when I'm doing a "real" application, I spent a lot of time doing quick toy implementations to test out different features along the way, probably more time than I spend on the "final" version of the script. So I believe making these toy implementations faster and easier has a potentially significant impact on total development time, even if not a single one of them ends up in the final "real application" script.

PR #415, opened 2024-03-26
Three bug fixes

daniel-klein:
1. Bernoulli (binomial) random variables don't use a random number when p=0, throwing off CRN. Small/hacky fix for that issue.
2. In demographics, better handling of age bins that was causing a bug in some cases.
3. Correct length of sim yearvec for dt != 1.

### DIFF
--- a/starsim/distributions.py
+++ b/starsim/distributions.py
@@ -11,7 +11,7 @@ from starsim.random import SingleRNG, MultiRNG
 from starsim import options
 from scipy.stats import (bernoulli, expon, lognorm, norm, poisson, randint, rv_discrete, 
                          uniform, rv_histogram, weibull_min)
-from scipy.stats._discrete_distns import bernoulli_gen # TODO: can we remove this?
+from scipy.stats._discrete_distns import bernoulli_gen, poisson_gen # TODO: can we remove this?
 
 
 __all__ = ['ScipyDistribution', 'ScipyHistogram']
--- a/starsim/distributions.py
+++ b/starsim/distributions.py
@@ -33,6 +33,27 @@ class ScipyDistribution():
                 self.context = context
                 return
 
+            def rvs_fix_bernoulli(self, *args, **kwargs):
+                # Bernoulli/binomial hack
+                # 0 in binomial (bernoulli) doesn't call rand, which throws off CRN!
+                # Solution 1: Replace 0 with eps
+                # Solution 2: Replace p where 0 with any non-zero value (e.g. eps), then revert vals back to 0
+                # --> Solution here is BOTH 1 and 2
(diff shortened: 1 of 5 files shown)


starsim/sim.py
```diff
@@ -189,7 +189,7 @@ def init_time_vecs(self):
         """
         Construct vectors things that keep track of time
         """
-        self.yearvec = np.arange(start=self.pars.start, stop=self.pars.end + 1, step=self.pars.dt)  # Includes all the timepoints in the last year
+        self.yearvec = np.arange(start=self.pars.start, stop=self.pars.end + self.pars.dt, step=self.pars.dt)  # Includes all the timepoints in the last year
```
cliffckerr, line 192: If we keep this change then minimally we should change the comment. But the reason it was originally written this way was so if results were computed annually regardless of timestep, then the last year would have the same number of timepoints as other years. Maybe we've decided not to do this, but in that case we should be careful that other assumptions in the code are consistent with this change.

tests/baseline.json
```diff
@@ -1,17 +1,17 @@
 {
   "summary": {
-    "n_alive": 9661.44761904762,
-    "new_deaths": 7.076190476190476,
-    "cum_deaths": 736.0,
+    "n_alive": 9673.058823529413,
```
cliffckerr, line 3: Why did the baseline change? Was it the change to yearvec, rvs_crn, or something else?

starsim/distributions.py
```diff
@@ -33,6 +33,36 @@ def initialize(self, sim, context):
...
+                    vals[inds] = 0
+                elif options.multirng and isinstance(self, poisson_gen) and ('lam' in kwargs) and not np.isscalar(kwargs['lam']):
+                    inds = np.where(kwargs['lam']==0)[0]
+                    kwargs['lam'][inds] = np.finfo(float).eps
+
+                    # Actually sample the random values
+                    vals = super().rvs(*args, **kwargs)
+
+                    # Complete the Poisson fix
+                    vals[inds] = 0
```
cliffckerr, line 59: I think this could be simplified into a single block of code, but perhaps not worth spending time on since hopefully will be removed once the `refactor-rngs` branch is merged!

daniel-klein: Agreed!

tests/baseline.json
daniel-klein: I didn't look to see if this change was driven by the difference in simulation duration or the bernoulli/poisson edge case, but could if you think it might be caused in error?

starsim/sim.py
daniel-klein: The simulation should end at the user-configured end! I will update the comment.

This bug was discovered in running a 6-month SIR simulation. Instead of stopping after 0.5 years, the simulation ran for 1.5 years - very strange. We should be aware that with the general purpose of the Starsim Framework in mind, simulation durations could vary from days to decades.

PR #420, opened 2024-03-28
Refactor rngs ppf

cliffckerr:
Refactors the RNG refactor to use slots, and to use `ppf` as an alternative solution to the repeated slot handling code.

SUPER ugly code, needs a LOT of tidying, but it seems like it works, maybe?!

In almost all cases, the "standard" solution works fine: take the basic method (e.g. `rng.uniform(low, high, size)`), where `size=n` if not using UIDs, and `size=sim.people.slot[uids].max()+1` if using UIDs.

The only case in which this _doesn't_ work is if there are repeated slots _and_ array (or callable) parameters. Now what happens in this case is that:
1. The number of random numbers drawn are as before: `size = sim.people.slot[uids].max()+1`
2. The _quantile_ for each slot is drawn: `qq = rng.random(sze)`
3. The slots in use are pulled out: `qqs = qq[slots]`
4. These are then mapped to the actual distribution via the inverse CDF, aka PPF, e.g.: `rvs = scipy.stats.uniform(low, high).ppf(qqs)`

Two agents who share the same slot will have exactly the same quantile, but they can still map to different values since the distribution parameter values can be different. Even though calling `ppf` is typically slower than generating new random numbers, since this is only for a corner case, and since `ppf` is only called for the slots that are in use, it should be as fast or faster than the other implementation.

For known issues, see https://github.com/starsimhub/starsim/issues/422


starsim/diseases/ncd.py
```diff
@@ -48,7 +48,7 @@ def set_initial_states(self, sim):
         alive_uids = ss.true(sim.people.alive)
         initial_risk = self.pars['initial_risk'].filter(alive_uids)
         self.at_risk[initial_risk] = True
-        self.ti_affected[initial_risk] = sim.ti + sc.randround(self.pars['dur_risk'].urvs(initial_risk) / sim.dt)
+        self.ti_affected[initial_risk] = sim.ti + sc.randround(self.pars['dur_risk'].rvs(initial_risk) / sim.dt)
```
daniel-klein, line 51: Note that randround is not CRN safe.

starsim/diseases/syphilis.py
```diff
@@ -263,7 +263,7 @@ def set_congenital(self, sim, target_uids, source_uids=None):
 
                 # Birth outcomes must be modified to add probability of susceptible birth
                 birth_outcomes = self.pars.birth_outcomes[state]
-                assigned_outcomes = birth_outcomes.urvs(uids)-uids  # WHY??
+                assigned_outcomes = birth_outcomes.rvs(len(uids))-uids  # WHY?? # TODO: TEMP: NOT CRN SAFE
```
daniel-klein, line 266: If there is a desire to make this code CRN safe, I think we could easily make `choice` CRN safe in this case (where the options are fixed). It's simply a roulette selection, each outcome has equal probability and use a single uniform random in inverse cumulative to select.

starsim/distributions.py
```diff
@@ -341,64 +353,99 @@ def make_dist(self, size, uids=None):
...
+            uids = None
+            slots = None
+            size = n
+        else:
+            uids = np.asarray(n)
+            if len(uids):
+                if self.slots is None:
+                    print('TEMP SLOTS')
+                    self.slots = np.arange(uids.max() + 1)
+                slots = np.asarray(self.slots[uids]) # TODO: check if asarray needed
```
daniel-klein, line 395: Seems like asarray shouldn't be needed, but would need to debug.

starsim/distributions.py
```diff
@@ -341,64 +353,99 @@ def make_dist(self, size, uids=None):
...
+        # Figure out size, UIDs, and slots
+        if np.isscalar(n) or isinstance(n, tuple):  # If passing a non-scalar size, interpret as dimension rather than UIDs iff a tuple
+            uids = None
+            slots = None
+            size = n
+        else:
+            uids = np.asarray(n)
+            if len(uids):
+                if self.slots is None:
+                    print('TEMP SLOTS')
```
daniel-klein, line 393: Remember to remove before final.

daniel-klein: Does the code ever hit this point in practice? The user provided UIDs, but we don't have slots. Might want to throw an exception rather than trying to continue with UIDs as slots?

starsim/distributions.py
```diff
@@ -341,64 +353,99 @@ def make_dist(self, size, uids=None):
...
         else:
-            raise ValueError(f'Unknown method: {self.method}') # Should not happen
+            rands = self.rand(size)[slots]
+            if self.method == 'numpy':
+                mapping = dict(normal='norm', lognormal='lognorm')
+                dname = mapping[self.dist] if self.dist in mapping else self.dist # TODO: refactor
+                if dname == 'uniform': # TODO: hack to get uniform to work since different args for SciPy
+                    kwds['loc'] = kwds.pop('low')
+                    kwds['scale'] = kwds.pop('high') - kwds['loc']
+                spdist = getattr(sps, dname)(**kwds) # TODO: make it work better, not actually numpy
```
daniel-klein, line 429: Do the kwds typically align between numpy and scipy implementations?

daniel-klein (commented):
Clever approach with the ppf. For some distributions, like normal and importantly bernoulli, we can adjust for array parameters _after_ sampling from the distribution. This could be a nice speed advantage by using built-in optimizations. And for other distributions, we'll have to resort to the ppf approach.

starsim/distributions.py
```diff
@@ -341,64 +353,99 @@ def make_dist(self, size, uids=None):
...
-    def urvs(self, uids):
-        """ Like rvs(), but get based on a list of unique identifiers (UIDs or slots) instead """
-        uids = np.asarray(uids)
-        if not len(uids):
-            return np.array([], dtype=int) # int dtype allows use as index, e.g. when filtering
-        maxval = uids.max() + 1 # Since UIDs are inclusive
-        urvs = self.rvs(size=maxval, uids=uids)
-        return urvs
 
     def filter(self, uids, **kwargs): # TODO: should this only be valid for Bernoulli distribution types?
```
daniel-klein, line 446: For now, I think it's safe to restrict to Bernoulli. One could later imagine fancy filtering, but we're not ready for that yet.

daniel-klein:
Note also that Bernoulli and Poisson need special handling as p=0 and lam=0, respectively, do not use random numbers.

starsim/distributions.py
cliffckerr: It hit it in testing only, but I've refactored the tests so this is no longer needed and has been removed.

starsim/distributions.py
cliffckerr: Not needed, removed

starsim/diseases/ncd.py
cliffckerr: Added to #422

starsim/distributions.py
cliffckerr: No -- will remove this hack when the supported distributions (e.g. `ss.weibull`) are subclassed. For unsupported distributions, the user can supply any distribution from `scipy.stats` (since these all have a `ppf` method), and then the user will supply the appropriate keywords for that distribution.

Subclassing part of #422

cliffckerr:
> Note also that Bernoulli and Poisson need special handling as p=0 and lam=0, respectively, do not use random numbers.

Would this still be an issue on this branch? If so could you write a test demonstrating it?

starsim/disease.py
```diff
@@ -287,9 +287,11 @@ def make_new_cases(self, sim):
 
                 slots_s = people.slot[src] # Slots for the possible source
                 slots_t = people.slot[trg] # Slots for the possible target
-                rvs_s = self.rng_source.rvs(size=np.max(slots_s) + 1)[slots_s]
-                rvs_t = self.rng_target.rvs(size=np.max(slots_t) + 1)[slots_t]
-                rvs = np.remainder(rvs_s + rvs_t, 1) # Generate a new random number based on the two other random numbers
+                slots = (slots_s + slots_t) % sim.pars.n_agents # TODO: is this ok?
+                rvs = self.rng_source.rvs(size=np.max(slots)+1)[slots]
```
daniel-klein, line 291: In short... no, not okay - many draws will get the same random number, and hence be correlated.

cliffckerr: There will be some pairs who get the same draws, just as there are with using slots. But I'd be curious to see a test that showed this actually caused an issue in practice!

cliffckerr: I'll revert this change in the mean time, but leave it commented so we can revisit it.

Issue #354, opened 2024-03-02
Refactor par_dists


daniel-klein:
Good discussion on slack resulting in a few options:

1. From @RomeshA:

```python
pars = ss.omergeleft(pars,
    dur_inf = 1,
    init_prev = 0.1,
    p_death = 0.2,
    beta = None,
)
        
pars.dur_inf = ss.make_dist(pars.dur_inf, ss.lognorm)
pars.init_prev = ss.make_dist(pars.init_prev, ss.bernoulli)
pars.p_death = ss.make_dist(pars.p_death, ss.bernoulli)
```

2. From @cliffckerr:
```python
pars = ss.omergeleft(pars,
    dur_inf = {'lognorm':1},
    init_prev = {'bernoulli':0.1},
    p_death = {'bernoulli':0.2},
    beta = None,
)
```

3. From me:
```python
pars = ss.omergeleft(pars,
    dur_inf = sps.lognorm(s=1)
    init_prev = 0.1, # Do not specify a distribution for Bernoulli, even though we will convert to a ScipyDistribution under the hood (in part to enable CRN and callable parameters).
    p_death = 0.2,
    beta = None,
)
```
```python
ss.set_distpars(disease.pars.dur_inf, s=5)
disease.pars.dur_inf.kwds['s'] = 5
disease.pars.dur_inf = sps.norm(loc=10)
```

cliffckerr:
Duplicate of https://github.com/starsimhub/starsim/issues/292

kaa2102:
`par_dists` usage in Starsim codebase

demographics.py
lines 115, 139, 153, 248, 273

network.py
lines 435, 530, 539, 550

modules.py
lines 74, 82

products.py
lines 143, 145,146

disease/cholera.py
lines 17, 42, 53

disease/ebola.py
lines 16, 38, 51

diseases/hiv.py
lines 14, 34, 39

diseases/gonorrhea.py
lines 13, 32, 39

disease/measles.py
lines 16, 30, 37

disease/sir.py
lines 21, 29, 35, 101, 119, 120, 181

docs/tutorials/tut_buildsim.ipynb*
line 106

cliffckerr:
Or we could do this:
```py
self.pars = dict(
    dur_inf = ss.lognorm_ex(1)
    init_prev = ss.bernoulli(0.1),
    p_death = ss.bernoulli(0.2),
    beta = None,
)
self.pars.update(**kwargs) # automatically merges with correct structure
```

[...]

PR #446, opened 2024-04-04
Remove bernoulli from par_dists

daniel-klein:
* Remove bernoulli from par_dists. There is no need to have bernoulli in `par_dists` as these _must_ be bernoulli. Hard coding to avoid potential mistakes.

* Removed duplicate states

* Lifted `init_prev` to Infection, which caused a stochastic change, so new baselines.

Related to #354

### DIFF
--- a/starsim/demographics.py
+++ b/starsim/demographics.py
@@ -147,10 +147,6 @@ class Deaths(Demographics):
             units = 1e-3,  # assumes death rates are per 1000. If using percentages, switch this to 1
         )
 
-        self.par_dists = ss.omergeleft(par_dists,
-            death_rate = ss.bernoulli
-        )
-
         # Process metadata. Defaults here are the labels used by UN data
         self.metadata = ss.omergeleft(metadata,
             data_cols = dict(year='Time', sex='Sex', age='AgeGrpStart', value='mx'),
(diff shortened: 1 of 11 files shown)


cliffckerr (commented):
I agree we should definitely remove this! This solution doesn't quite work, as it prevents the user from providing a Starsim dist:
```py
import starsim as ss

sim = ss.Sim(pars=dict(
    diseases = dict(
        type = 'sir',
        p_death = ss.bernoulli(0.1),
        ),
    networks = 'random',
    )
)

sim.run()

TypeError: Dist.__call__() takes from 1 to 2 positional arguments but 4 were given
```

But I 100% agree with you about `par_dists` and that Bernoulli distributions shouldn't be substituted for other things! Will propose a fix in a PR hopefully very soon.

daniel-klein:
Thanks for the review, @cliffckerr. I intentionally chose to prevent users from providing their own distribution. It has to be Bernoulli, and the only thing a user should have to think about is the probability. Here, the user can provide a scalar, to be used for all agents, or a callable function, which will be handled by the RNG logic. But keen to see if you have a better way to go about this.

cliffckerr:
Closed by #488

Issue #447, opened 2024-04-04
Allow age_data values to handle ints


skroiss-idm:
An error is thrown if the 'values' column in age_data are of datatype int64.

kaa2102:
Are we considering (1) changing the data type or (2) going upstream to enable the class parameters to accept int64?

daniel-klein:
Perhaps related to #451?

devclinton:
@skroiss-idm  - Is this fixed with #451?

skroiss-idm:
Ahh, yes. Looks like that's fixed! Thanks for flagging.

PR #451, opened 2024-04-08
Fix type conversion in age_props

daniel-klein:
Trivial one-line change to avoid a warning (or error?) in converting int to float.


daniel-klein:
The underlying issue this PR addresses is the following error message:
```python
numpy.core._exceptions._UFuncOutputCastingError: Cannot cast ufunc 'divide' output from dtype('float64') to dtype('int64') with casting rule 'same_kind'
```

daniel-klein:
Perhaps related to #447?

PR #455, opened 2024-04-11
Fix ti vs year comparisons

daniel-klein:
Several disease modules compared a time index (ti) with `sim.year`, which is the absolute year like 2022.63. This PR reverts everything back to time indices, but there's a case for shifting everything to floating point time (year) instead of discrete indices. Let's debate, but perhaps merge this PR first considering it should resolve issues with deaths.


cliffckerr:
Baselines aer currently only computed for SIR and HIV -- not for all diseases. Would be good to think about coverage specifically for baseline!

Issue #423, opened 2024-03-28
Rethink `UIDArray.__getitem__()`


cliffckerr:
Currently, it takes the vast majority of time in `make_new_cases()`:
![image](https://github.com/starsimhub/starsim/assets/3239256/c7d801cc-03df-47a2-b5c5-314b0a207ae4)
(See https://github.com/starsimhub/starsim/issues/422 for more performance tests.)

Is there a way we could use more standard Numpy operations to make this faster, or reduce some of the checking? It's already Numbafied, so can't get any additional speed increases there.

Another option would be to go back to not removing agents, so the UID is also just the array index.

cliffckerr:
Another benchmarking:
```py
import sciris as sc
import starsim as ss

with sc.cprofile(sort='selfpct') as cpr:
    sim = ss.demo(n_agents=100e3, start=2000, end=2100, diseases='sis', plot=False)

sim.diseases.sis.plot()
```
gives
```
                                                  func    cumpct     selfpct  cumtime    selftime  calls                                         path
0                                       _get_vals_uids   35.5618  3.5518e+01   1.5462  1.5443e+00   1317                                 states.py:83
1    <method 'permutation' of 'numpy.random._genera...   12.5981  1.2597e+01   0.5478  5.4770e-01    102                                            ~
2                                       make_new_cases   61.5934  8.5633e+00   2.6780  3.7233e-01    101                               disease.py:255
3                                      __array_ufunc__    8.2585  8.1880e+00   0.3591  3.5601e-01   3950                                states.py:267
4                                                  rvs   27.3114  4.3405e+00   1.1875  1.8872e-01    509                         distributions.py:440
5                                            add_pairs   19.5998  4.1161e+00   0.8522  1.7897e-01    102                               network.py:497
6    <method 'lognormal' of 'numpy.random._generato...    3.0262  3.0262e+00   0.1316  1.3158e-01    102                                            ~
7                                          beta_per_dt    2.3906  2.3694e+00   0.1039  1.0302e-01    202                               network.py:288
8    <method 'random' of 'numpy.random._generator.G...    2.0724  2.0724e+00   0.0901  9.0107e-02    406                                            ~
9           <method 'reduce' of 'numpy.ufunc' objects>    1.6476  1.6434e+00   0.0716  7.1452e-02    821                                            ~
10                                           end_pairs    6.6629  1.1970e+00   0.2897  5.2043e-02    101                               network.py:324
```

cliffckerr:
[...]

Issue #407, opened 2024-03-21
State subclasses


cliffckerr:
Instead of a single `ss.State()`, I think it would be clearer to have subclasses that were specific for each  type, e.g. bool, int, float, since it would be an error to substitute one for the other (e.g. define `susceptible` as a float). I also personally feel that `State` makes sense only for bool arrays (`alive` is a state, but `rel_sus` to me is more of a property/parameter). So I suggest that instead of (from `disease.py`):
```py
ss.State('susceptible', bool, True),
ss.State('infected', bool, False),
ss.State('rel_sus', float, 1.0),
ss.State('rel_sev', float, 1.0),
ss.State('rel_trans', float, 1.0),
ss.State('ti_infected', int, ss.INT_NAN),
```
we have
```py
ss.BoolArr('susceptible', True), # no default value for boolean arrays
ss.BoolArr('infected', False),
ss.FloatArr('rel_sus', 1.0), # np.nan would be the default fill value
ss.FloatArr('rel_sev', 1.0),
ss.FloatArr('rel_trans', 1.0),
ss.IntArr('ti_infected'), # ss.INT_NAN would be the default fill value so doesn't need to be specified
```

If users want other dtypes, they could use `ss.State` directly (which I would probably suggest renaming `ss.Arr`).

This would also have the advantage that we could optimize computations. For example, if we use `np.int32` as the dtype for `IntArr` (which I propose we should!), and we use `np.int32()` on the inputs, then we should get a 2x speedup on the ufunc arithmetic. We could still do this from within a single `State` class, but I think it would be less clear, and then the onus would be on the user to remember to do `ss.State('ti_infected', np.int32, ss.INT_NAN)` everywhere, which is getting ugly.

daniel-klein:
This seems like a good place to use options in settings.py. We can autodetect the data type consistently, so it seems to be a question of precision, right?

cliffckerr:
Closed by #456

Issue #400, opened 2024-03-20
Make INT_NAN negative


cliffckerr:
When doing operations on arrays containing small numbers of INT_NAN, the results can be slightly larger than expected, which is confusing. I think if INT_NAN were negative it would be clearer that something has gone wrong! (But also, we should probably make methods either exclude INT_NAN by default, or return `np.nan`, rather than include some extraneous very large numbers.)

See also #421

cliffckerr:
Closed by #456

cliffckerr:
Closed by #456

Issue #304, opened 2024-02-23
Defined and undefined should work with INT_NAN


robynstuart:

cliffckerr:
Fixed

PR #456, opened 2024-04-13
Rethink UIDArray

### LINKED ISSUE #296: Check that dead-but-not-removed agents are being handled sensibly
if we implement this change, wouldn't we need to go through all the modules and make sure that we were always conditioning on people being alive before selecting them for e.g. pair formation, disease acquisition/progression, etc?

_Originally posted by @robynstuart in https://github.com/amath-idm/starsim/issues/294#issuecomment-1958601482_

cliffckerr:
*Update 2024-04-23:* **This PR is now ready to review + merge!**

Replaces `UIDArray`, `ArrayView`, and `State` with `Arr`. 

### Dead agents
Dead agents are no longer removed, so `uid` always corresponds to the position in the array. This means that no remapping is necessary, which has a significant performance benefit; running this script:
```py
import sciris as sc
import starsim as ss

kw = dict(n_agents=100e3, start=2000, end=2100, diseases='sir', plot=False)

with sc.cprofile(sort='selfpct', mintime=1e-1) as cpr:
    sim = ss.demo(**kw)
```
for me takes about 2.7 s on `main` and about 1.8 s on this branch.

Instead, each `Arr` (formerly known as `State`) carries around a copy of the `People` object, from which it can access `aliveinds`, which is used to filter the arrays by default (only if UIDs are _not_ provided: if they are, they can be used without any extra filtering). When an agent dies, all of their properties are set to `nan` (see below), and their UID is removed from `aliveinds`.

### What else has changed?
`Arr` has type-specific subclasses, like `FloatArr` and `BoolArr`. These classes have sensible defaults defined, so e.g. `ss.State('ti_dead', int, ss.INT_NAN)` can be replaced with `ss.IntArr('ti_dead')`.

These classes also have a defined `nan` value (for float and int, not for boolean). These also have only appropriate operations defined on them, e.g. on `main`:
```py
>>> sim = ss.demo()
>>> sim.diseases.sir.ti_infected
      Quantity
UID           
0            1
1            2
2            1
3            2
4            2
...        ...
9995         2
9996         2
9997         2
9998         2
9999         0

[9806 rows x 1 columns]
>>> ~sim.diseases.sir.ti_infected # uh oh, bitwise inverse of an int
      Quantity
UID           
0           -2
1           -3
2           -2
3           -3
4           -3
...        ...
9995        -3
9996        -3
9997        -3
9998        -3
9999        -1

[9806 rows x 1 columns]
```
On this branch:
```py
>>> sim = ss.demo()
>>> sim.diseases.sir.ti_infected
<IntArr "ti_infected", len=4064, [1 2 1 ... 2 2 0]>
>>> ~sim.diseases.sir.ti_infected

  Cell In[2], line 1
    ~sim.diseases.sir.ti_infected

  File ~/idm/starsim/starsim/states.py:119 in __invert__
    def __invert__(self):     raise BooleanOperationError(self)

BooleanOperationError: Logical operations are only valid on Boolean arrays, not <class 'numpy.int64'>
```

### Outstanding tasks

Other than fixing the failing tests, a major outstanding question is how chaining should work. For performance and simplicity, an operation on an `Arr` returns a `np.array`. However, in some cases, it's necessary to do operations on an `Arr` before indexing it with `uids`. We could just do the operation on the whole array (rather than just `aliveinds`), but this is inefficient. 

The solution is to define a `people.remap_uids()` function that will "expand" the sparse array (of length `len(people.aliveinds)`) back into the underlying array (of length `max(people.uid)`). This is pretty fast, and is actually only used in one place currently (`make_new_cases`), but is ugly. 

Closes #296, #304, #400, #407, #423


starsim/people.py
```diff
@@ -68,60 +73,41 @@ def register_state(self, state, die=True):
...
-        self.slot.grow(new_uids)
-        self.slot[new_uids] = new_slots if new_slots is not None else new_uids
+        new_slots = new_slots if new_slots is not None else new_uids
+        self.slot.grow(new_uids, new_vals=new_slots)
 
         for state in self._states.values():
             state.grow(new_uids)
+            
+        # Finally, update the alive indices
+        self.aliveinds = np.concatenate([self.aliveinds, new_uids], axis=0)
```
RomeshA, line 107: I think this has to happen earlier so that the new UIDs aren't filtered out when updating other `Arr` instances

starsim/people.py
```diff
@@ -68,60 +73,41 @@ def register_state(self, state, die=True):
...
-        new_inds = np.arange(start_idx, start_idx + n)
-
-        self._uid_map.grow(n)
-        self._uid_map[new_uids] = new_inds
-
-        self.uid.grow(n)
-        self.uid[new_inds] = new_uids
+        start_uid = self.uid.len_used
+        stop_uid = start_uid + n
+        new_uids = np.arange(start_uid, stop_uid)
```
RomeshA, line 96: This causes an error because of the indexing where `np.arange` on my machine returns an `int32` so it gets interpreted as indexing `Arr.values` rather than `Arr._arr`. I suspect changing these behaviours depending on integer type won't be robust enough

starsim/people.py
```diff
@@ -20,38 +20,43 @@ class BasePeople(sc.prettyobj):
...
     def __len__(self):
         """ Length of people """
-        return len(self.uid)
+        return len(self.aliveinds)
+    
+    @property
+    def n_uids(self):
+        return self.uid.len_used
+    
+    def remap_uids(self, arr):
```
RomeshA, line 55: I think this could be an `Arr` method so that users could request a full array from `Arr` on demand. It could be more efficient to just return a slice of `Arr._arr` up to `n_uids`? However, it would be confusing if that returns a reference whereas `.values` returns a copy

RomeshA (commented):
@cliffckerr I think my main question (also for everyone) would be about if there will be performance consequences of dropping the dead agent removal functionality in general - we originally planned to add this to Starsim following performance issues in HPVsim for applications where the number of dead agents was very large, so my concern would be that the tests used to compare performance on this branch thus far don't cover that scenario. We should at least run some sims that are more representative of those HPVsim-type runs to see what to expect.

It would also be a bit of a shame to lose the ability to filter down the UIDs and so go back towards the days of having to keep track of arrays of indices to filter things down. That could be mitigated a bit if the people filtering in this PR was more generalized, at the moment it's hardcoded to be `aliveinds` but presumably it could be something different

CHANGELOG.rst
```diff
@@ -7,6 +7,13 @@ What's new
 All notable changes to the codebase are documented in this file. Changes that may result in differences in model output, or are required in order to run an old parameter set with the current version, are flagged with the term "Regression information".
 
 
+Version 0.4.0 (2024-04-13)
+--------------------------
+- Replace ``UIDArray``, ``ArrayView``, and ``State`` with ``Arr``.
```
kaa2102, line 12: I think we can add a note that `State` is replaced with {data.type}Arr, e.g., BoolArr, IntArr, etc.

starsim/diseases/measles.py
kaa2102, line 1: 1. Was numpy not needed in `measles.py` before?
2. What is the rationale for switching `ss.State('ti.exposed', float,np.nan)` from data type **_Float_** to **_Int_** `(ss.IntArr('ti.exposed'))`?

starsim/diseases/sir.py
kaa2102, line 1: Curious to know the performance difference in using ss.true() vs sc.findinds()?

starsim/utils.py
kaa2102, line 1: Great way to simplify the code!

kaa2102 (approved):

daniel-klein:
Strong +1 to @RomeshA's main point about more extensive performance testing regarding the utility of removing agents. How about a 100 year simulation with vital dynamics resembling a geography where Starsim is/will be applied? Keen to compare performance of the current `State`-based approach with and without removing dead agents to this new approach. Thanks @cliffckerr!

cliffckerr:
Thank you for your reviews @RomeshA , @kaa2102 , and @daniel-klein ! All good points and will respond to them when I get a chance, which is likely to be next week. A quick note about performance: the number of elements in an array doesn't affect the time required to index it (only the number of indices). In this implementation, either a finite set of `uids` is being pulled out, or, if no UIDs are provided, then `aliveinds` is used to index. In either case, the dead agents should be skipped over with no performance penalty (other than memory).

The reason HPVsim was so slow with dead agents was because of the proliferation of multiscale zombie ghost agents, and because the boolean array calculations did _not_ skip over dead agents. Those issues shouldn't apply here, but definitely good to test the latter!

starsim/diseases/cholera.py
```diff
@@ -56,15 +56,15 @@ def __init__(self, pars=None, par_dists=None, *args, **kwargs):
...
     
             # Timepoint states
-            ss.State('ti_exposed', float, np.nan),
-            ss.State('ti_symptomatic', float, np.nan),
-            ss.State('ti_recovered', float, np.nan),
-            ss.State('ti_dead', float, np.nan),
+            ss.FloatArr('ti_exposed'),
+            ss.FloatArr('ti_symptomatic'),
+            ss.FloatArr('ti_recovered'),
+            ss.FloatArr('ti_dead'),
```
robynstuart, line 67: All the timepoints should be IntArr

Actually, I almost feel like these could be added automatically during initialization -- for all BoolArr states, we add a corresponding ti_ IntArr state. Possibly only for those that are False by default.

starsim/diseases/ebola.py
```diff
@@ -54,17 +54,17 @@ def __init__(self, pars=None, par_dists=None, *args, **kwargs):
...
-            ss.State('ti_exposed', float, np.nan),
-            ss.State('ti_severe', float, np.nan),
-            ss.State('ti_recovered', float, np.nan),
-            ss.State('ti_dead', float, np.nan),
-            ss.State('ti_buried', float, np.nan),
+            ss.FloatArr('ti_exposed'),
+            ss.FloatArr('ti_severe'),
+            ss.FloatArr('ti_recovered'),
+            ss.FloatArr('ti_dead'),
+            ss.FloatArr('ti_buried'),
```
robynstuart, line 67: should all be IntArr

starsim/people.py
```diff
@@ -22,39 +22,44 @@ class BasePeople(sc.prettyobj):
...
-        self.uid = ss.ArrayView(int, default=ss.INT_NAN)  # This variable tracks all UIDs currently in use
-
         n = int(n_agents)
-
-        self._uid_map.grow(n)
-        self._uid_map[:] = np.arange(0, n)
-        self.uid.grow(n)
-        self.uid[:] = np.arange(0, n)
+        self.initialized = False
+        self.uid = ss.IndexArr('uid')  # This variable tracks all UIDs currently in use
```
robynstuart, line 27: 'in use' meaning alive?

starsim/settings.py
```diff
@@ -9,9 +9,9 @@
 import numpy as np
 import sciris as sc
 
-__all__ = ['INT_NAN', 'dtypes', 'options']
+__all__ = ['intnan', 'dtypes', 'options']
 
-INT_NAN = 2147483647 # From np.iinfo(np.int32).max: value to use to flag invalid content (i.e., an integer value we are treating like NaN, since NaN can't be stored in an integer array)
+intnan = -32767 # From np.iinfo(np.int16).max: value to use to flag invalid content (i.e., an integer value we are treating like NaN, since NaN can't be stored in an integer array)
```
robynstuart, line 14: maybe a dumb question, but why couldn't this just be -1?

starsim/utils.py
```diff
@@ -241,63 +246,18 @@ def set_seed_regular(seed):
...
+#     """
+#     return np.nonzero(arr)[0]
 
-    **Example**::
-
-        inds = ss.false(people.alive) # Returns array of UIDs of dead agents
-    """
-    return _false(state.uid.__array__(), state.__array__())
 
 def defined(arr):
```
robynstuart, line 262: will these work? defined/undefined? are they used anywhere?

robynstuart (approved):
Overall this looks great, but I'd love to see an additional test script or demo script indicating how the API is supposed to be used.

starsim/diseases/cholera.py
cliffckerr: Agree, I want to define a ``ss.Transition()`` object, kind of like Dymodetron, that would automatically set up the states, state transitions, and time of transition.

cliffckerr: Fixed here though

starsim/diseases/ebola.py
cliffckerr: Fixed

starsim/people.py
cliffckerr: clarified

starsim/settings.py
cliffckerr: it could be, but one can imagine situations where a state could store a negative integer -- e.g. for a disease where symptoms could appear either before or infectiousness, a variable tracking the difference could end up being positive or negative. Then -1 day would be erroneously interpreted as a NaN. Another reason is that in an array with even a few of these NaN values, if they're not removed, they make the results look sufficiently weird that it's clear something's gone wrong!

starsim/utils.py
cliffckerr: no, removed!

cliffckerr:
> Overall this looks great, but I'd love to see an additional test script or demo script indicating how the API is supposed to be used.

Added to `test_base.py`

RomeshA:
@cliffckerr In the process of going over the changes and going to try it out with the HIV module we're currently implementing, I did just encounter some odd behaviour though?

```
self.partners.values
Out[13]: array([-32767, -32767, -32767, ..., -32767, -32767, -32767], dtype=int64)
self.concurrency.values
Out[14]: array([1, 1, 1, ..., 1, 1, 1], dtype=int64)
self.partners.values < self.concurrency.values
Out[15]: array([ True,  True,  True, ...,  True,  True,  True])
self.partners < self.concurrency
Out[16]: <BoolArr "partners", len=10001, [False False False ... False False False]>
```

starsim/states.py
```diff
@@ -24,462 +25,193 @@ def check_dtype(dtype, default=None):
...
-                values, uids, new_uid_map = gvu_func(self.values, np.fromiter(key, dtype=int), self._uid_map.__array__())
-            return UIDArray(values=values, uid=uids, uid_map=new_uid_map)
-        except IndexError as e:
-            if str(INT_NAN) in str(e):
-                raise IndexError(f'UID {key} not present in array')
-            else:
-                raise e
-
+        if isinstance(key, uids): # Check that it's UIDs # TODO: think about slice, list, etc
+            return self.raw[key]
```
RomeshA, line 92: This one is failing if the UIDs are empty e.g.,

```
sim.people.age[[]]
Out[6]: array([], dtype=float64)
ss.uids([])
Out[7]: uids([], dtype=float64)
sim.people.age[ss.uids([])]
  File "C:\Users\romesh\miniconda3\envs\atomica312\Lib\site-packages\IPython\core\interactiveshell.py", line 3577, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-8-2d09ba477783>", line 1, in <module>
    sim.people.age[ss.uids([])]
    ~~~~~~~~~~~~~~^^^^^^^^^^^^^
  File "C:\Users\romesh\projects\idm\starsim\starsim\states.py", line 92, in __getitem__
    return self.raw[key]
           ~~~~~~~~^^^^^
IndexError: arrays used as indices must be of integer (or boolean) type
```

starsim/states.py
```diff
@@ -501,59 +233,142 @@ def initialize(self, sim):
...
+        super().__init__(name=name, dtype=ss_bool, default=default, nan=nan, label=label, coerce=False, skip_init=skip_init)
         return
+    
+    def __and__(self, other): return self.asnew(self.values & other)
+    def __or__(self, other):  return self.asnew(self.values | other)
+    def __xor__(self, other): return self.asnew(self.values ^ other)
+    def __invert__(self):     return self.asnew(~self.values)
+    
+    @property
+    def uids(self):
```
RomeshA, line 324: I like having a method to do this rather than `ss.true` but can we call the method `true` instead of `uids` and have a corresponding `false` method as well? That would also then allow `auids` to be renamed `uids`

starsim/states.py
```diff
@@ -24,462 +25,193 @@ def check_dtype(dtype, default=None):
...
-            if str(INT_NAN) in str(e):
-                raise IndexError(f'UID {key} not present in array')
-            else:
-                raise e
-
+        if isinstance(key, ss.BoolArr):
+            key = key.uids
+            
+        if isinstance(key, (int, uids, slice)):
+            self.raw[key] = value
```
RomeshA, line 104: Same issue as in `__getitem__` raising an error with an empty array (`ss.uids([])`). This was happening a bit when trying to convert the HIV code over where we unconditionally read/write from the arrays with a collection of UIDs but if nobody is sampled on that timestep then that collection is empty.

starsim/states.py
```diff
@@ -24,462 +25,193 @@ def check_dtype(dtype, default=None):
...
-        except IndexError as e:
-            if str(INT_NAN) in str(e):
-                raise IndexError(f'UID {key} not present in array')
-            else:
-                raise e
-
+        if isinstance(key, ss.BoolArr):
+            key = key.uids
+            
+        if isinstance(key, (int, uids, slice)):
```
RomeshA, line 103: I was wondering about this not being symmetric with `__getitem__` with respect to slices - here, the slice acts on `.raw` whereas in `__getitem__` the slice acts on `.values`. This can lead to some weird-looking behaviour once some of the agents have died e.g.,

```
sim.people.age[0:5]
Out[14]: array([22.08333333, 42.08333333, 64.08333333, 28.08333333, 33.08333333])
sim.people.age[0:5] = 1
sim.people.age[0:5]
Out[16]: array([ 1.        ,  1.        ,  1.        , 28.08333333, 33.08333333])
```

starsim/states.py
```diff
@@ -24,462 +25,193 @@ def check_dtype(dtype, default=None):
...
-            if kv >= len(uid_map):
-                raise IndexError('UID not present in array (requested UID is larger than the maximum UID in use)')
-            idx = uid_map[kv]
-            if idx == INT_NAN:
-                raise IndexError('UID not present in array')
-            elif idx >= len(vals):
-                raise Exception('Attempted to write to a non-existant index - this can happen if attempting to write to new entries that have not yet been allocated via grow()')
-            vals[idx] = value
-
     def __getitem__(self, key):
```
RomeshA, line 87: I was expecting that this would work

```
sim.people.ti_dead[sim.people.uid]
  File "C:\Users\romesh\miniconda3\envs\atomica312\Lib\site-packages\IPython\core\interactiveshell.py", line 3577, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-29-e17a9156e088>", line 1, in <module>
    sim.people.ti_dead[sim.people.uid]
    ~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^
  File "C:\Users\romesh\projects\idm\starsim\starsim\states.py", line 97, in __getitem__
    return self.values[key]
           ~~~~~~~~~~~^^^^^
IndexError: index 11794 is out of bounds for axis 0 with size 11793
```

but since `IndexArr` is not a `uids` instance it does something different? If you put in a scalar UID type it could be used to here to have `IndexArr` also do lookups by UID

starsim/settings.py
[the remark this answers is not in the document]
cliffckerr: I don't think a formal one; I used lowercase to be consistent with `np.nan`. In general, Numpy doesn't use uppercase for constants (e.g. `np.pi`), nor does SciPy.

For internal-use constants (e.g. `np.lib.format.BUFFER_SIZE`), I agree we should use the convention of uppercase.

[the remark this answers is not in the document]
cliffckerr: `self.ti_dead <= sim.ti` still works because of the `self.notnan()` call. It _won't_ work to do `self.ti_dead.raw <= sim.ti`, though (see comment below).

I agree that for int arrays, which tend to be time indices, we use `<` much more often than `>`. I see two conflicting arguments here:
1. We should use a very large value to represent NaN, so if a mistake is made and it isn't getting filtered out as it should, it isn't likely to matter.
2. We should use a negative value to represent NaN, so mistakes are less likely to go unnoticed and thus get fixed.

Ironically I feel I'm usually on the side of (1) and you're usually on the side of (2), but it seems we're swapped here :) 

My personal reason for preferring a negative value is perhaps no longer relevant, but it's that a few times when I was looking at the arrays (especially derived quantities like `self.ti_dead.mean()`), a few large positive values mixed in with many "correct" values didn't immediately jump out to me as an error. To me negative values more clearly indicate that something's not right (and I do like @robynstuart's suggestion of just using -1, but that feels restrictive for a general `IntArr` class).

I had a look at pandas' [nullable integer series](https://pandas.pydata.org/docs/user_guide/integer_na.html) -- it looks like exactly what we need, but I think performance is considerably worse than plain Numpy arrays.

starsim/network.py
```diff
@@ -71,7 +71,7 @@ def __init__(self, pars=None, key_dict=None, vertical=False, *args, **kwargs):
             p2 = ss_int_,
```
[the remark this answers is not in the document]
cliffckerr: You can use a single scalar value:
```py
import starsim as ss
sim = ss.demo(plot=False)
sim.people.age[ss.uids(9999)]
126.41934324595512 # need to fix the age bug lol
```

We could implement it this way, I prefer having a `ss.uids` class to support other methods, e.g. I like being able to do `self.auids = self.auids.remove(uids)`. But there is an annoyance here, namely that because `ss.uids` is implemented as a view of an array, you can't modify the array in-place (because it doesn't own its own data, or there are multiple references to the data). If we could get around this by subclassing int instead, that might be preferable!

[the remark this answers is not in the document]
RomeshA: Could `self.auids.remove(uids)` be implemented as `Arr.remove()` (and same for intersect) to provide similar convenience? I think having the notion of a 'UID' being a thing (via subclassing int) will be clearer than having the UID-ness be a property of a collection of ints

starsim/settings.py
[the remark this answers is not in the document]
RomeshA: I don't think we should rely too much on the `notnan()` call because indexing the `IntArr` turns it back into a normal array, right? So not just using `.raw` but if they were operating on a collection of UIDs that would also be the same right (like `self.ti_dead[uids] <= sim.ti`)? 

My thinking originally was that just writing `<= sim.ti` would be the intended use case and not a mistake, hence it should 'just work' - and aside from that, it wasn't obvious to me whether a large positive or a large negative value would be more noticeable as a mistake in general. Probably context dependent although I'd agree that in typical `ti_*` usages a negative value would be more noticeable. The magnitude also matters, I suspect `-32767` might be too small, if the user is taking a sum/average it might not be big enough to outweigh valid entries and the aggregate might still look OK, so I think whichever sign we go with, a larger magnitude would decrease (but not eliminate) the chances of that happening.

cliffckerr:
> @cliffckerr In the process of going over the changes and going to try it out with the HIV module we're currently implementing, I did just encounter some odd behaviour though?
> 
> ```
> self.partners.values
> Out[13]: array([-32767, -32767, -32767, ..., -32767, -32767, -32767], dtype=int64)
> self.concurrency.values
> Out[14]: array([1, 1, 1, ..., 1, 1, 1], dtype=int64)
> self.partners.values < self.concurrency.values
> Out[15]: array([ True,  True,  True, ...,  True,  True,  True])
> self.partners < self.concurrency
> Out[16]: <BoolArr "partners", len=10001, [False False False ... False False False]>
> ```

I think this should be fixed now, but not sure how to reproduce it so not 100% sure!

starsim/settings.py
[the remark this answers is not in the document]
cliffckerr: Removed in latest revision

starsim/network.py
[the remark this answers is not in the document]
cliffckerr: I think we could implement it this way, but I prefer the current implementation for a couple reasons:
- It matches how `pd.Series` is implemented, e.g.:
```py
s = pd.Series([1,2,None,4], dtype='Int64')
type(s[0])
Out[10]: numpy.int64 # but this is not the Series dtype, which is pd.types.Int64
```
- This way it's easy to check the type (`isinstance(key, ss.uids)`), whereas otherwise it gets complicated. In particular, Numpy seems to jam int-like things into `np.int64`, e.g.
```py
import numpy as np

class UID(np.int64):
    pass

a = np.arange(5, dtype=UID)
assert a.dtype == UID # passes
assert a.dtype != np.int64 # fails
AssertionError
```

starsim/states.py
```diff
@@ -24,462 +25,193 @@ def check_dtype(dtype, default=None):
...
-            return UIDArray(values=values, uid=uids, uid_map=new_uid_map)
-        except IndexError as e:
-            if str(INT_NAN) in str(e):
-                raise IndexError(f'UID {key} not present in array')
-            else:
-                raise e
-
+        if isinstance(key, uids): # Check that it's UIDs # TODO: think about slice, list, etc
+            return self.raw[key]
+        elif isinstance(key, np.ndarray) and key.dtype == np.int64:
```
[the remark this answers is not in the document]
cliffckerr: Fixed

starsim/states.py
```diff
@@ -24,462 +25,193 @@ def check_dtype(dtype, default=None):
...
-            if isinstance(out, UIDArray):
-                # For some operations (e.g., those involving two UIDArrays) the result of the ufunc will already be a UIDArray
-                # In particular, operating on two states will result in a UIDArray where the references to the original People uid_map and uids
-                # are still intact. In such cases, we can return the resulting UIDArray directly
-                return out
+    def values(self):
+        return self.raw[self.auids] # TODO: think about if this makes sense for uids
+        
+    def isnan(self):
+        errormsg = f'Arr.isnan() not implemented for {self}'
```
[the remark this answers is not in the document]
cliffckerr: Fixed

starsim/states.py
```diff
@@ -24,462 +25,193 @@ def check_dtype(dtype, default=None):
...
-        return self.values.any(*args, **kwargs)
-
-    def all(self, *args, **kwargs):
-        return self.values.all(*args, **kwargs)
-
-    def count(self, *args, **kwargs):
-        return np.count_nonzero(self.values, *args, **kwargs)
-
+        
+    def __gt__(self, other): return self.asnew(self.notnan(self.values > other),  cls=BoolArr)
```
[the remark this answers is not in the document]
cliffckerr: Fixed

starsim/states.py
cliffckerr: Fixed

starsim/states.py
cliffckerr: The rationale for this implementation is:
- `Arr.auids` maps directly to `sim.people.auids` (which is an array of active UID indices), which we need to distinguish from `sim.people.uid` (which is an Arr containing all UIDs ever created)
- I thought it could get confusing which methods return UIDs/inds and which return booleans, which is why I thought `.uids` clarified that it was returning a list of UIDs in a way that `true()` wouldn't
- The ufuncs should be working, so `arr.false()` would here be `~arr.uids`. I like this better because it clarifies that you can invert an `Arr` and use it as an `Arr` (`~arr`), or you can also just pull out the False UIDs (`~arr.uids`). With `false()`, it always just returned the UIDs rather than giving this option.

starsim/states.py
cliffckerr: Fixed

starsim/states.py
cliffckerr: Fixed

starsim/states.py
cliffckerr: Fixed

starsim/interventions.py
```diff
@@ -198,9 +204,9 @@ def __init__(self, product=None, prob=None, eligibility=None, **kwargs):
         self.prob = sc.promotetoarray(prob)
         self.eligibility = eligibility
         self._parse_product(product)
-        self.screened = ss.State('screened', bool, False)
-        self.screens = ss.State('screens', int, 0)
-        self.ti_screened = ss.State('ti_screened', int, ss.INT_NAN)
+        self.screened = ss.BoolArr('screened')
+        self.screens = ss.FloatArr('screens', default=0)
```
daniel-klein, line 208: Seems liek `screens` should be an `IntArr`?

starsim/interventions.py
```diff
@@ -198,9 +204,9 @@ def __init__(self, product=None, prob=None, eligibility=None, **kwargs):
         self.prob = sc.promotetoarray(prob)
         self.eligibility = eligibility
         self._parse_product(product)
-        self.screened = ss.State('screened', bool, False)
-        self.screens = ss.State('screens', int, 0)
-        self.ti_screened = ss.State('ti_screened', int, ss.INT_NAN)
+        self.screened = ss.BoolArr('screened')
+        self.screens = ss.FloatArr('screens', default=0)
+        self.ti_screened = ss.FloatArr('ti_screened')
```
daniel-klein, line 209: Also I would expect that `ti_*` are `IntArray` as well.

starsim/network.py
```diff
@@ -780,30 +788,28 @@ def add_pairs(self, people, ti=None):
                 print('No pairs to add')
             return 0
 
-        available = np.concatenate((available_m, available_f))
+        available = ss.uids.cat(available_m, available_f)
         loc = self.pars.embedding_func.rvs(available)
+        if np.isnan(loc).sum():
+            raise Exception('WARNING, should not happen!')
+            loc = sc.rmnans(loc, replacenans=9999)
```
daniel-klein, line 795: Looks like temporary code?

starsim/network.py
```diff
@@ -780,30 +788,28 @@ def add_pairs(self, people, ti=None):
...
 
         n_pairs = len(ind_f)
 
-        beta = np.ones(n_pairs)
-
-        # Figure out durations
+        # Finalize pairs
         p1 = available_m[ind_m]
+        p2 = available_f[ind_f]
+        beta = np.ones(n_pairs) # TODO: fix
```
daniel-klein, line 808: What remains TODO / fix here?

starsim/states.py
RomeshA: I think I was coming at it with the notion that an `Arr` corresponded to a subset of the UIDs, and then the ambiguity as to whether `BoolArr.uids` was the UIDs contained in the `Arr`, or the UIDs for which it was true. I was expecting `uids` to be the former, although maybe that's just something that people need to learn (to use `auids` for that). 

But I wonder if that's making too much of `Arr.auids` mapping directly to `sim.people.auids` - it would be cool eventually to be able to filter the `People` in different ways by having a different mask on the `Arr` instance, in which case `Arr.auids` might _not_ match `sim.people.auids`. That's the scenario I was thinking of in having `Arr.uids` mean 'the UIDs that appear in this `Arr`' and `Arr.true`/`Arr.false` clearly referring to the true/false entries. 

I don't think that `true` and `false` return the UIDs would be too confusing though, since that's what `cv.true()` and `cv.false()` did (and what `ss.true()` and `ss.false()` currently do)?

starsim/interventions.py
cliffckerr: Discussed on the call -- no more `IntArr` due to `np.nan` support.

starsim/interventions.py
cliffckerr: Discussed on the call -- no more `IntArr` due to `np.nan` support.

starsim/network.py
cliffckerr: Removed

starsim/network.py
cliffckerr: Updated the comment -- would be nice to potentially have an ability to set beta for new connections. This ends up being confusing with beta-change interventions that must get reapplied at each timestep to modify new connections, rather than having a way to just set the default beta for new connections. I'm imagining something like `beta = np.full(n_pairs, self.beta)`.

PR #481, opened 2024-04-24
Add pull request template

### LINKED ISSUE #471: Add pull request template
Add checklist as reminder to update version and changelog

cliffckerr:
### Description
Adds PR template, closes #471

### Checklist
- [ ] A new version number is needed & changelog has been updated

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -3,6 +3,6 @@
 
 ### Checklist
 - [ ] Code commented & docstrings added
-- [ ] A new version number was needed & changelog has been updated
 - [ ] New tests were needed and have been added
+- [ ] A new version number was needed & changelog has been updated
 - [ ] A new PyPI version needs to be released
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -2,7 +2,7 @@
 
 
 ### Checklist
-- [ ] Code is appropriately commented
-- [ ] A new version number is needed & changelog has been updated
+- [ ] Code commented & docstrings added if needed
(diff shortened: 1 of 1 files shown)


kaa2102 (approved):
Good to go!

RomeshA:
For a checklist is the idea that all of the items are checked off as a prerequisite for merging? That would make it easy to quickly see that everything had been covered. In that case, something like "New tests were needed and have been added" could instead be "New tests not required OR relevant tests have been added" (or similar) so that way a user that didn't need new tests would still check the box. Same with the new version number e.g. "Version number has been incremented and changelog updated OR new version number not required"

kaa2102:
> For a checklist is the idea that all of the items are checked off as a prerequisite for merging? That would make it easy to quickly see that everything had been covered. In that case, something like "New tests were needed and have been added" could instead be "New tests not required OR relevant tests have been added" (or similar) so that way a user that didn't need new tests would still check the box. Same with the new version number e.g. "Version number has been incremented and changelog updated OR new version number not required"

I agree with this sentiment. In discussions, I mentioned that version and changelog updates may be required after a significant change or 4-5 merges. The checkbox option @RomeshA mentioned should be available.

cliffckerr:
I went back and forth on this (as you can see from the commit history!), but decided on this approach because I didn't want it to be a, well, box-ticking exercise! I assume that when making the PR, having the boxes be there will be enough of a reminder to the PR author. And then during PR review, I thought _not_ having all the boxes necessarily ticked would be more informative for the reviewer (e.g. whether they agree with the author that new tests aren't required). But happy to revert to the "added or not needed" version if everyone prefers that!

Issue #440, opened 2024-04-02
Fix SIR plotting


cliffckerr:
Uses timestep rather than sim step -- would be fixed if modules had access to the Sim object.

daniel-klein:
As far as I know, SIR is unique in having a `plot()` method as part of the disease module. Seems like a reasonable pattern, but does not extend well to multi-disease simulations etc. In addition to`module.plot()`, we previously discussed a pattern like `sim.plot(modules=sir)`.

cliffckerr:
Closed by #488

Issue #412, opened 2024-03-22
Default sim ages


RomeshA:
I was running a test sim using `ss.Sim()` directly (with no arguments) and was puzzled why my simulation had nobody younger than 50, and people that were up to 150 years old. Eventually I realised this is just because the default age distribution is 0-100 and the simulation is run for 50 years. I think this might be a bit confusing for users to encounter in a 'default' sim though?

daniel-klein:
Agree!
```python
dist = ss.uniform(low=0, high=100, name='Age distribution')
```
... is not a particularly useful default.

cliffckerr:
Closed by #488

daniel-klein:
The default distribution if no age data is provided remains:
```python
dist = ss.uniform(low=0, high=100, name='Age distribution')
```
This issue was closed by #488, which says, "The default maximum age if none is specified is 50 instead of 100." However it's still 100 in code.

Issue #178, opened 2023-12-11
Update msm and fix mf_msm connector


daniel-klein:
mf_msm connector code is out of date an no longer functional. Also needs a test, not just a devtest.

cliffckerr:
Closed by #488

PR #488, opened 2024-05-04
Rethink Sim API

### LINKED ISSUE #301: Rethink Starsim parameters and API
There are lots of quirks and inconveniences with how parameters are supplied, initialized, and validated. For example, this works:
```py
sim = ss.Sim(pars=dict(networks='random', diseases='sir'), people=ppl, demographics=demographics)
```

But these should also work:

```py
# Supply strings as kwargs
sim = ss.Sim(people=ppl, demographics=demographics, networks='random', diseases='sir')


# Supply people as a parameter
pars=dict(networks='random', diseases='sir', people=ppl, demographics=demographics)
sim = ss.Sim(pars)
```

Tasks:
- [x] Move initialization into `parameters.py`
- [x] Replace `dictmergeleft()` with `pars.update()`
- [x] Replace `par_dists` (was #354)
- [x] Reimplement `convert_plugins` (was #284)
- [x] Copy inputs flag (was #103)
- [x] Allow smarter dict-based parameter input
- [x] Link modules to sim

cliffckerr:
#### Summary
All inputs to the sim and modules now use a `ss.Pars()` class, which handles updating and validation. It is now not necessary to ever use `pars=` (although you still can if you want), so what was previously:
`sim = ss.Sim(pars=dict(diseases='sir', networks='random'))`
is now just:
`sim = ss.Sim(diseases='sir', networks='random')`

Updates happen recursively, so distributions etc. can be flexibly updated.

This has significantly changed how modules are initialized; what was previously

```py
def __init__(self, pars=None, **kwargs):

    pars = ss.omergeleft(pars,
        dur_inf = 6,
        init_prev = 0.01,
        p_death = 0.01,
        beta = 0.5,
    )

    par_dists = ss.omergeleft(par_dists,
        dur_inf = ss.lognorm_ex,
        init_prev = ss.bernoulli,
        p_death = ss.bernoulli,
    )

    super().__init__(pars=pars, par_dists=par_dists, *args, **kwargs)
```

is now:
```py
def __init__(self, pars=None, **kwargs):
    super().__init__()
    self.default_pars(
        beta = 0.5,
        init_prev = ss.bernoulli(0.01),
        dur_inf = ss.lognorm_ex(6),
        p_death = ss.bernoulli(0.01),
    )
    self.update_pars(pars, **kwargs)
```

#### Parameter changes
- Added a `ss.Pars` class (and a `ss.SimPars` subclass) that handles parameter creation, updates, and validation.
- Initialization has been moved from `sim.py` to `parameters.py`; `ss.Sim.convert_plugins()` has been replaced by `ss.SimPars.convert_modules()`.
- The key method is `ss.Pars.update()`, which performs all necessary validation on the parameters being updated.

#### Module changes
- Whereas modules previously initialized a dict of parameters and then called `super().__init__(pars, **kwargs)`, they now call `super().__init__()` first, then `self.default_pars(par1=x, par2=y)`, then finally `self.update_pars(pars, **kwargs)`.
- What was previously e.g. `ss.Module(pars=dict(par=x))` is now `ss.Module(par=x)`.
- `par_dists` has been removed; instead, distributions are specified in the default parameters, and are updated via the `Pars` object.
- `ss.module_map()` maps different module types to their location in the sim.
- `ss.find_modules()` finds all available modules (including subclasses) in Starsim.
- Removed `ss.dictmerge()` and `ss.dictmergeleft` (now handled by `ss.Pars.update()`).
- Removed `ss.get_subclasses()` and `ss.all_subclasses()` (now handled by `ss.find_modules()`).
- Modules now contain a link back to the `Sim` object.
- Added `to_json()` and `plot()` methods to `Module`.
- Removed `connectors.py`; connectors still exist but as an empty subclass of `Module`.

#### People and network changes
- Time parameters (`ti`, `dt`, etc.) have been removed from `People`. Use `sim.ti`, `sim.dt` etc. instead. One consequence of this is that `people.request_death()` now requires a `sim` argument. Another is that network methods (e.g. `add_pairs()`) now take `sim` arguments instead of `people` arguments.
- `SexualNetwork` is now a subclass of `DynamicNetwork`.
- Removed `ss.Networks` (now just an `ss.ndict`).
- Network connectors have been removed.
- `Person` has been implemented as a slice of `sim.people[i]`.
- The default maximum age if none is specified is 50 instead of 100.
- Agents do not age if no demographics modules are supplied.

#### Other changes
- All inputs to a sim are now copied by default. To disable, use `ss.Sim(..., copy_inputs=False)`.
- There is a new `Plugin` class, which contains shared logic for Interventions and Analyzers. It has a `from_func()`, which will generate an intervention/analyzer from a function.
- Diseases no longer have a default value of `beta=1` assigned; beta must be defined explicitly if being used.
- Individual diseases can now be plotted via either e.g. `sim.plot('hiv')` or `sim.diseases.hiv.plot()`.
- Distributions can be created from dicts via `ss.make_dist()`.
- A new function `ss.check_sims_match()` will check if the results of two or more simulations match.
- Merged `test_dcp.py` and `test_base.py` into `test_other.py`.
- Renamed `test_simple.py` to `test_sim.py`.
- Renamed `test_dists.py` to `test_randomness.py`.

### Checklist

Closes #301, #178, #412, #440


CHANGELOG.rst
```diff
@@ -7,6 +7,101 @@ What's new
...
+- Added ``to_json()`` and ``plot()`` methods to ``Module``.
+- Removed ``connectors.py``; connectors still exist but as an empty subclass of ``Module``.
+
+People and network changes
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+- Time parameters (``ti``, ``dt``, etc.) have been removed from ``People``. Use ``sim.ti``, ``sim.dt`` etc. instead. One consequence of this is that ``people.request_death()`` now requires a ``sim`` argument. Another is that network methods (e.g. ``add_pairs()``) now take ``sim`` arguments instead of ``people`` arguments.
+- ``SexualNetwork`` is now a subclass of ``DynamicNetwork``.
+- Removed ``ss.Networks`` (now just an ``ss.ndict``).
+- Network connectors have been removed.
+- ``Person`` has been implemented as a slice of ``sim.people[i]``.
```
kaa2102, line 87: Cool! I like the simplifications with type, time steps, SexualNetwork, person -> sim.people[i], et. al.

docs/tutorials/tut_interventions.ipynb
```diff
@@ -40,12 +40,13 @@
...
    "source": [
     "import starsim as ss\n",
+    "\n",
     "pars = dict(\n",
     "    n_agents = 5_000,\n",
     "    birth_rate = 20,\n",
     "    death_rate = 15,\n",
     "    networks = dict(\n",
-    "        type = 'randomnet',\n",
+    "        type = 'randomnet', # Or 'random'\n",
```
kaa2102, line 49: Will both 'randomnet' and 'random' work?

starsim/diseases/ebola.py
```diff
@@ -13,45 +13,30 @@
...
+        self.default_pars(
             # Initial conditions and beta
-            init_prev = 0.005,
-            beta = None,
-            sev_factor = 2.2,
+            init_prev       = ss.bernoulli(0.005),
+            beta            = 1.0, # Placeholder value
+            sev_factor      = 2.2,
             unburied_factor = 2.1,
+
```
kaa2102, line 25: Python standards implemented for the win! Good stuff!

docs/tutorials/tut_interventions.ipynb
cliffckerr: Yes, either works!

starsim/parameters.py
```diff
@@ -2,85 +2,431 @@
...
+    
+    Acts like an ``sc.objdict()``, except that adding new keys are disallowed by
+    default, and auto-updates known types.
+    """
+    def __init__(self, pars=None, **kwargs):
+        if pars is not None:
+            if isinstance(pars, dict):
+                kwargs = pars | kwargs
+            else:
+                errormsg = f'Cannot supply parameters as type {type(pars)}: must be dict'
```
menriquez-IDM, line 29: you could have used:
raise ValueError(f'Cannot supply parameters as type {type(pars)}: must be a dict')

starsim/parameters.py
```diff
@@ -2,85 +2,431 @@
...
+                raise ValueError(errormsg)
+        super().__init__(**kwargs)
+        return
+    
+    def update(self, pars=None, create=False, **kwargs):
```
menriquez-IDM, line 34: Hi Cliff, this function is great, and I think it will add a great value to the framework, also, while reading it, I noticed that readability may be a challenge with the way how it's been written so I was wonderinf if it could be refactored in smaller parts, only to improve readability and manteinibility, this is what I am referring to (this is jus a suggestion, and has not be tested)...

```python
    pars = self._prepare_pars(pars, **kwargs)
    if not pars: 
        return self

    if not create:
        self.check_key_mismatch(pars)
    
    for key, new in pars.items():
        if key not in self.keys(): 
            self[key] = new
        else: 
            self._update_key(key, new, create)

    return self

def _prepare_pars(self, pars, **kwargs):
    pars = {} if pars is None else dict(pars)
    return pars | kwargs

def _update_key(self, key, new, create):
    old = self[key]
    if isinstance(old, atomic_classes):
        self[key] = new
    elif isinstance(old, Pars): 
        old.update(new, create=create)
    elif isinstance(old, (ss.ndict, ss.Module)):
        self._update_container(old, new)
    elif isinstance(old, ss.Dist):
        self._update_distribution(old, new)
    else:
        raise TypeError(f'No known mechanism for handling {type(old)} → {type(new)}; using default')

def _update_container(self, old, new):
    if not len(old): 
        self[key] = new
    elif isinstance(new, dict):
        for newkey, newvals in new.items():
            old[newkey].pars.update(newvals)
    else:
        raise TypeError(f'Cannot update an {type(old)} with {type(new)}: must be a dict to set new parameters')

def _update_distribution(self, old, new):
    if isinstance(new, ss.Dist): 
        self._update_dist_with_dist(old, new)
    elif isinstance(new, Number):
        old.set(new)
    elif isinstance(new, list):
        old.set(*new)
    elif isinstance(new, dict):
        self._update_dist_with_dict(old, new)
    else:
        raise TypeError(f'Cannot update a distribution with {type(new)}')

def _update_dist_with_dist(self, old, new):
    if isinstance(old, ss.bernoulli) and not isinstance(new, ss.bernoulli):
        raise TypeError(f"Bernoulli distributions can't be changed to another type: {type(new)} is invalid")
    else:
        self[key] = new

def _update_dist_with_dict(self, old, new):
    if 'type' not in new.keys(): 
        old.set(**new)
    else: 
        self._create_new_distribution(old, new)

def _create_new_distribution(self, old, new):
    newtype = new['type']
    if isinstance(old, ss.bernoulli) and newtype != 'bernoulli':
        raise TypeError(f"Bernoulli distributions can't be changed to another type: {newtype} is invalid")
    else:
        dist = ss.make_dist(new)
        self[key] = dist
 ```

File: starsim/parameters.py [hunk 6]
```diff
@@ -2,85 +2,431 @@
...
 
+__all__ = ['Pars', 'SimPars', 'make_pars']
 
-__all__ = ['Parameters', 'make_pars']
+# Define classes to not descend into further -- based on sciris.sc_nested
+atomic_classes = (str, Number, list, np.ndarray, pd.Series, pd.DataFrame, type(None))
 
-class Parameters(sc.objdict):
+
+class Pars(sc.objdict):
```
menriquez-IDM, line 17: Not sure if there is a strong reason to call the parameters "par / pars" but since it's getting its own class, could it be possible to get it fixed? 

> "par" 
> Abbreviation of paragraph
> Abbreviation of parenthesis
> Abbreviation of parish
> Also, par =
 _"parallel"_
> 
> "param" is the actual abreviation for parameters. 

I know it may be a huge change because this is everywhere, but I wanted to share my POV.

starsim/parameters.py
cliffckerr: Fixed, thx!

starsim/parameters.py
cliffckerr: I agree `Params` or `Parameters` would be less ambiguous. However, we have a loose convention of instances being named the same as classes except for the capitalized first letter where possible (e.g. `obj = Obj()`), and in this case we have `module.pars = ss.Pars()`, and we use "pars" to mean parameters in lots of other contexts (e.g. `export_pars()`, `update_pars()`). I think it would reduce readability and conciseness if we had `self.params` everywhere instead of `self.pars`, especially given that I think it's clear in this context that "pars" can't mean paragraphs, parentheses, or parishes! But I agree it's a balance. I had `self.p` for a while but this was a little short even for my taste, and `module.p = ss.P()` is not a line of code anyone could love :D

starsim/parameters.py
cliffckerr: I like this @menriquez-IDM! I've refactored it along those lines.

starsim/distributions.py
```diff
@@ -461,6 +488,8 @@ def rvs(self, n=1, reset=False):
         # Check for readiness
         if not self.initialized:
             raise DistNotInitializedError(self)
+        # elif self.initialized == 'partial':
+        #     print(f'WARNING: {self} called on {self.sim.ti} with {n}')
```
daniel-klein, line 492: Remove

tests/test_diseases.py
```diff
@@ -39,27 +39,92 @@ def test_sir():
...
+        sim.plot()
+        sim.diseases.sis.plot()
+    return sim
+
+
 def test_ncd():
     ppl = ss.People(n_agents)
     ncd = ss.NCD()
-    sim = ss.Sim(people=ppl, diseases=ncd)
+    sim = ss.Sim(people=ppl, diseases=ncd, copy_inputs=False) # Since using ncd directly below
```
daniel-klein, line 127: Why is `copy_inputs` is needed here?

CHANGELOG.rst
```diff
@@ -7,6 +7,101 @@ What's new
...
+- The default maximum age if none is specified is 50 instead of 100.
```
daniel-klein, line 88: What are the implications of a default max age of 50?

CHANGELOG.rst
```diff
@@ -7,6 +7,101 @@ What's new
...
+- Agents do not age if no demographics modules are supplied.
```
daniel-klein, line 89: Wait, really?! Instead I would recommend a sim parameter `enable_aging` that would likely default to True, even without demographic modules.

tests/test_sim.py
```diff
@@ -0,0 +1,181 @@
...
+def make_sim_pars():
+    pars = sc.objdict(
+        n_agents = n_agents,
+        birth_rate = 20,
+        death_rate = 15,
+        networks = sc.objdict(
+            type = 'randomnet',
+            n_contacts = 4,
+        ),
+        diseases = sc.objdict(
```
daniel-klein, line 26: With the new API, how would a user configure multiple networks or disease modules? If it is possible, add a test here.

tests/baseline.json
```diff
@@ -1,17 +1,18 @@
 {
   "summary": {
-    "n_alive": 9647.450980392157,
-    "new_deaths": 7.490196078431373,
-    "cum_deaths": 755.0,
+    "yearvec": 2010.1000000000026,
```
daniel-klein, line 3: `yearvec` in summary? Is this at the final time point?

tests/devtests/devtest_connectors.py
```diff
@@ -4,31 +4,128 @@
...
+            syphilis.susceptible[treat_ids] = True
+            sim.diseases.hiv.rel_sus[treat_ids] = 1
+            sim.diseases.hiv.rel_trans[treat_ids] = 1
+
+
+# Make people, HIV, syphilis, and network
+def make_args():
+    # Marital
+    mf = ss.MFNet(
+        pars = dict(
```
daniel-klein, line 74: With these API changes, is it still necessary to put module parameters in a `dict`?

daniel-klein (commented):
Nice updates. I still prefer the detailed per-module approach, as I expect that to be the dominant usage pattern, but appreciate the alternatives. With the simplified dict-based configuration approach, I'm not clear if it is possible for a user to configure and parameterize multiple modules (disease, network, interventions...).

daniel-klein:
One additional API suggestion, perhaps more closely related to states, is regarding accessing uids from booleans. Previously, we had
```python
uids = ss.true(condition)
```
and now it is
```python
uids = (condition).uids
```

I'd prefer function syntax here, something like:
```python
uids = ss.uids(condition)
```
... realizing that `ss.uids` is likely taken already. Whatever it's called, it could simply return `.uids`.

cliffckerr:
The new test that was added that's failing is fixed by #500

starsim/parameters.py
```diff
@@ -2,85 +2,355 @@
...
+            if key not in self.keys(): # It's a new parameter and create=True: update directly
+                self[key] = new
+            else:
+                old = self[key] # Get the existing object we're about to update
+                if isinstance(old, atomic_classes): # It's a number, string, etc: update directly
+                    self[key] = new
+                elif isinstance(old, Pars): # It's a Pars object: update recursively
+                    old.update(new, create=create)
+                elif isinstance(old, ss.ndict): # Update module containers
+                    self._process_ndict(key, old, new)
```
pausz, line 62: Since these methods (`self._process_*()`) are updating the parameters of a given starsim class, could we simply call them `update_pars_ndict` or `update_ndict_pars`, `update_module_pars`? 

"Process" gives me the impression that the method is doing something more than updating parameters.

starsim/parameters.py
```diff
@@ -2,85 +2,431 @@
...
+        self.end        = None   # End of the simulation
+        self.n_years    = 50     # Number of years to run, if end isn't specified. Note that this includes burn-in
+        self.dt         = 1.0    # Timestep
+        self.rand_seed  = 1      # Random seed, if None, don't reset
+        self.slot_scale = 5      # Random slots will be assigned to newborn agents between min=n_agents and max=slot_scale*n_agents
 
         # Demographic parameters
-        self.location    = None  #  NOT CURRENTLY FUNCTIONAL - what demographics to use
+        self.location   = None  #  NOT CURRENTLY FUNCTIONAL - what demographics to use
         self.birth_rate = None
```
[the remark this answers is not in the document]
pausz: I think it's useful to have birth_rates and death_rates here and users can set a numeric value. It's easy and accessible. I see this case similar to the cases where a parameter can be a single number ( or constant distribution) or specified via a distribution. 

In the case of birth_rates and death_rates, it'd be nice if we could specify them as 

- (1) a single number (or constant temporal pattern, equivalent to constant distribution of parameters that are expected to be random variables) 
- (2) a user-defined function (or pre-defined temporal pattern, like a sinusoid pattern to represent an observed uptick in birth rates for instance)
- (3) as a module that implements a more complex model of natality (mortality) and outputs the instantaneous birth rate (death rate).


I realise it probably needs to be solved in another PR.

starsim/parameters.py
```diff
@@ -2,85 +2,355 @@
...
+        """ Redefine default dict.update(), since overwritten in this class; should not usually be used """
+        super().update(*args, **kwargs)
+        return
+    
+    def to_json(self, **kwargs):
+        """ Convert to JSON representation """
+        return sc.jsonify(self, **kwargs)
+
+
+class SimPars(Pars):
```
pausz, line 147: The typical use case described in the docstring encapsulates the parameter creation within the ss.Sim() instantiation. While this is convenient, it doesn't make it obvious that this is the expected way to access Starsim's
default parameters. I think a typical use case involves users wanting to craft, manipulate, or inspect the parameters independently before passing them to a simulation. 


Perhaps we could name it `StarsimPars` or `DefaultPars`, it could give a better hint about the content and its scope. Because almost every starsim class has its own parameters, calling it `SimPars` could give the impression that these are parameters specific, and exclusive, to the Sim class itself, rather than defining more generally the model or scenario to simulate.

starsim/parameters.py
```diff
@@ -2,85 +2,355 @@
...
+                            modtype = modtype.lower() # Because our map is in lowercase
+                            moddictkeys = ssmoddict.keys()
+                            moddictvals = ssmoddict.values()
+                            if modtype in moddictkeys:
+                                modcls = ssmoddict[modtype]
+                            else:
+                                errormsg = f'Invalid module name "{modtype}" for "{modkey}"; must be one of {moddictkeys}'
+                                raise TypeError(errormsg)
+                        else: # Allow supplying directly as a class, e.g. dict(type=ss.SIR, dur_inf=6)
+                            if modtype in moddictvals:
```
pausz, line 316: This line will fail? If `moddictvals` is defined, then this code does not execute, but if this line executes, then moddictvals won't be defined; or is moddictvals defined earlier and i missed it :thinking:?

starsim/parameters.py
```diff
@@ -2,85 +2,355 @@
...
-        self.analyzers = ss.ndict()
+        self.analyzers     = ss.ndict()
 
         # Update with any supplied parameter values and generate things that need to be generated
         self.update(kwargs)
+        return
 
-        if self.slot_scale < 1:
-            raise Exception('The value of the "slot_scale" parameter must be a number >= 1.0')
+    def init_modules(self, sim, reset=False, **kwargs):
```
pausz, line 195: `sim`, `reset` and `kwargs` are no longer needed by this function   

Also, while parameter validation methods probably have found their way home, i think this `init` and init demographics method should in `Sim`. It's the job of the Sim class to orchestrate the flow of initialisation.  Also from the pov of an user who is adding a new module, I think it'd be confusing to find the initialisation int he parameter class.  

I've pushed some additional refactoring -- I thought it'd be quicker than trying to explain in a comment, but revert them if you don't agree with them .

starsim/people.py
```diff
@@ -272,11 +259,51 @@ def scale_flows(self, inds):
         """
         return self.scale[inds].sum()
 
+    def update_post(self, sim):
+        """ Final updates at the very end of the timestep """
+        if sim.demographics: # Only update ages if demographics are specified
+            self.age[self.alive.uids] += sim.dt
+        return
+
+    def request_death(self, sim, uids):
```
pausz, line 268: Is it really necessary to pass the sim object here? 
Can we just pass sim.ti to this method, and the argument be something like `tod` (time of death) or `tod_i` (to indicate is an index) 

 ```python 
def request_death(self, tod_i,  uids):
 ```

starsim/people.py
```diff
@@ -272,11 +259,51 @@ def scale_flows(self, inds):
...
+
+        :param uids: Agent IDs to request deaths for
+        :return: UIDs of agents that have been scheduled to die on this timestep
+        """
+        self.ti_dead[uids] = sim.ti
+        return
+
+    def resolve_deaths(self):
+        """ Carry out any deaths that took place this timestep """
+        death_uids = (self.ti_dead <= self.sim.ti).uids
```
pausz, line 298: I have noticed we're perhaps missing a potential opportunity to enhance the  "testability" of our code -- though there's a great test suite already. 

For instance this method:

```python
def resolve_deaths(self): 
    death_uids = (self.ti_dead <= self.sim.ti).uids
```
relies on the initialized `sim` object within people.  This could, however, impact the isolated testing of the method.

What about we try decoupling this method by, for instance,  injecting `current_ti` (or a shorter name my brain cannot imagine :grin:)  as an argument of the method:

```python
def resolve_deaths(self, current_ti): 
    death_uids = (self.ti_dead <= current_ti).uids
```

In this way, the method rely solely on its required input arguments, making
its functionality more transparent, simpler to understand and perhaps if we apply this approach to other methods, it'll make easier to test and debug the whole People class.

pausz (approved):

starsim/parameters.py
cliffckerr: Fixed in c2daea6

starsim/parameters.py
cliffckerr: My thinking was that these _are_ parameters that are specific to the `Sim` class -- while these do (often) contain other parameters, it's only in the `Sim` class that it would make sense to define e.g. `diseases` or `n_years` or `verbose` as an input. And they are only the default parameters for the sim -- the default parameters for, say, SIR are set by the `SIR.default_pars()` method. Have left this as-is for now.

starsim/parameters.py
cliffckerr: Fixed, moved definition earlier

starsim/parameters.py
cliffckerr: Fixed, thanks for your changes on this!

starsim/people.py
cliffckerr: It's not, this has been fixed

starsim/people.py
cliffckerr: I agree that this would make testing easier, but I feel paradoxically it would make it more likely that things would break and need to get caught by the tests :D Since right now `sim.ti` is the only sensible value for `current_ti`, having this be an argument requires the user to pass it correctly up the chain, which could lead to subtle bugs (e.g. if at some point they accidentally type `sim.dt` instead of `sim.ti`, which wouldn't cause an exception).

starsim/diseases/hiv.py
```diff
@@ -11,35 +11,30 @@
...
+            ss.FloatArr('ti_art'),
+            ss.FloatArr('ti_dead'), # Time of HIV-cause death
+            ss.FloatArr('cd4', default=500),
         )
 
-        super().__init__(pars=pars, par_dists=par_dists, *args, **kwargs)
         self.death_prob_data = sc.dcp(self.pars.death_prob)
-        self.pars.death_prob = self.make_death_prob
-
+        self.pars.death_prob = ss.bernoulli(self.make_death_prob)
```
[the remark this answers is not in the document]
cliffckerr: Fixed

starsim/diseases/sir.py
```diff
@@ -17,22 +17,15 @@ class SIR(ss.Infection):
...
-        pars = ss.dictmergeleft(pars,
-            dur_inf = 6,
-            init_prev = 0.01,
-            p_death = 0.01,
+    def __init__(self, pars=None, **kwargs):
+        super().__init__()
+        self.default_pars(
             beta = 0.5,
+            init_prev = ss.bernoulli(0.01),
+            dur_inf = ss.lognorm_ex(6),
```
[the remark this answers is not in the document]
cliffckerr: Fixed

starsim/demographics.py
```diff
@@ -249,9 +246,19 @@ def finalize(self, sim):
...
-        super().__init__(pars, **kwargs)
-
+    def __init__(self, pars=None, metadata=None, **kwargs):
+        super().__init__()
+        self.default_pars(
+            dur_pregnancy = 0.75,
+            dur_postpartum = 0.5,
+            fertility_rate = 0, # See make_fertility_prob_function
+            rel_fertility = 1,
+            maternal_death_rate = ss.bernoulli(0),
```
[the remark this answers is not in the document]
cliffckerr: Fixed

starsim/demographics.py
```diff
@@ -249,9 +246,19 @@ def finalize(self, sim):
...
 
-    def __init__(self, pars=None, par_dists=None, metadata=None, **kwargs):
-        super().__init__(pars, **kwargs)
-
+    def __init__(self, pars=None, metadata=None, **kwargs):
+        super().__init__()
+        self.default_pars(
+            dur_pregnancy = 0.75,
+            dur_postpartum = 0.5,
+            fertility_rate = 0, # See make_fertility_prob_function
```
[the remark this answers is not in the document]
cliffckerr: Agree, we need to think about how to handle this more elegantly.

starsim/network.py
```diff
@@ -585,40 +560,31 @@ def get_contacts(self):
...
-    def __init__(self, pars=None, par_dists=None, key_dict=None, **kwargs):
-        pars = ss.dictmergeleft(pars,
-            duration = 15,  # Can vary by age, year, and individual pair. Set scale=exp(mu) and s=sigma where mu,sigma are of the underlying normal distribution.
-            participation = 0.9,  # Probability of participating in this network - can vary by individual properties (age, sex, ...) using callable parameter values
-            debut = 16,  # Age of debut can vary by using callable parameter values
-            acts = 80,
+    def __init__(self, pars=None, key_dict=None, **kwargs):
+        super().__init__(key_dict=key_dict)
+        self.default_pars(
+            duration = ss.lognorm_ex(15),  # Can vary by age, year, and individual pair. Set scale=exp(mu) and s=sigma where mu,sigma are of the underlying normal distribution.
```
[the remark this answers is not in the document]
cliffckerr: Fixed

starsim/network.py
```diff
@@ -762,15 +727,16 @@ def __init__(self, pars=None, **kwargs):
...
+            male_shift is the average age that males are older than females in partnerships
         """
-        pars = ss.dictmerge({
-            'embedding_func': ss.normal(name='EmbeddingNet', loc=self.embedding_loc, scale=2),
-            'male_shift': 5,
-        }, pars)
-        super().__init__(pars, **kwargs)
+        super().__init__()
+        self.default_pars(
+            inherit = True, # The MFNet already comes with pars, we want to keep those
```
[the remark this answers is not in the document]
cliffckerr: I don't think that would work, since the module won't be in a valid state without the parent parameters being inherited. They can always be overwritten by user-supplied values, however.

starsim/parameters.py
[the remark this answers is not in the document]
cliffckerr: Agree -- no need to remove functionality since it's only a few lines of code to implement.

starsim/parameters.py
```diff
@@ -2,85 +2,431 @@
...
+                        errormsg = f'Was expecting {modkey} entry {i} to be class {expected_cls}, but was {type(mod)} instead'
+                        raise TypeError(errormsg)
+                    modlist[i] = mod
+                
+        return
+
+    def init_demographics(self):
+        """ Initialize demographics """
+
+        # Allow users to add vital dynamics by entering birth_rate and death_rate parameters directly to the sim
```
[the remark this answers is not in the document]
cliffckerr: I think we do -- see conversation above.

starsim/parameters.py
```diff
@@ -2,85 +2,431 @@
...
+            dt = self.dt
+            reciprocal = 1.0 / dt  # Compute the reciprocal of dt
+            if not reciprocal.is_integer():  # Check if reciprocal is not a whole (integer) number
+                # Round the reciprocal
+                reciprocal = int(reciprocal)
+                rounded_dt = 1.0 / reciprocal
+                self.dt = rounded_dt
+                if self.verbose:
+                    warnmsg = f"Warning: Provided time step dt: {dt} resulted in a non-integer number of steps per year. Rounded to {rounded_dt}."
+                    print(warnmsg)
```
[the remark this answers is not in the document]
cliffckerr: I think this is ok -- if the unit isn't a year, don't worry about trying to validate this. In any case, this will all be getting refactored very soon!

starsim/distributions.py
cliffckerr: Fixed -- actually completely redid how `self.initialized = 'partial'` is used

CHANGELOG.rst
cliffckerr: None, really. But I agree this is arbitrary in probably not a good way :) Will revert.

CHANGELOG.rst
cliffckerr: Added -- it defaults to `True` with demographics modules and `False` otherwise, but can be overridden.

tests/test_sim.py
cliffckerr: I added a new test, `test_sim.py:test_complex_api`, which illustrates both cases. As far as I can tell, every possible way of configuring a sim with objects is possible with dicts too (at least to two levels of depth).

tests/baseline.json
cliffckerr: Yes, this is showing the mean value. Which is a bit odd-looking, I agree, but it does convey useful information (e.g. if we change the definition of `start` and `end`, which might happen soon!).

tests/devtests/devtest_connectors.py
cliffckerr: No, have refactored this devtest and promoted it to a regular test!

cliffckerr:
@daniel-klein I made it so you can do `ss.uids(BoolArr)` instead of `BoolArr.uids` ... I would personally discourage this usage, since `ss.uids()` isn't valid on anything _other_ than a `BoolArr`, so I think the property makes it clearer that tis related specifically to that array. In particular, the previous implementation of `ss.true()` would just give indices, which would match UIDs if and only if (a) the input array matched the length the raw data, or (b) the user then did a subsequent transformation. In practice one of these two things was usually true, but sometimes not, leading to subtle bugs! In contrast, with `arr.uids`, it's guaranteed to be correct.

PR #498, opened 2024-05-10
Multiple interventions using the same product

alina-muellenmeister:
I ran into issues with multiple interventions using the same product. For each intervention using that product, it was previously trying to initialize the product and add it to the people class. 
I added a `def initialize(self, sim)` method to the `ss.Product` class to ensure the product only gets initialized once.
And I moved `self.people.add_module(intervention)` to the module initialization and removed it from `sim.py`. This way, the module is responsible for adding itself to the people class.

### DIFF
--- a/starsim/modules.py
+++ b/starsim/modules.py
@@ -108,6 +108,10 @@ class Module(sc.quickobj):
             state.initialize(sim)
 
         self.initialized = True
+
+        # Add module states to the People's dicts
+        sim.people.add_module(self)
+        
         return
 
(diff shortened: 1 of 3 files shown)


cliffckerr:
Thank you @alina-muellenmeister! Will merge this after #488 and #500

cliffckerr (approved):
Will need to reconcile with the other big PRs, but this looks like a helpful fix -- thank you!

alina-muellenmeister:
Thanks, Cliff!

Issue #497, opened 2024-05-10
Demographic results should be handled like other modules


daniel-klein:
I'm writing a custom demographics module based on Pregnancy. A custom intervention averts pregnancy-related deaths. However, there's no way to record true results (after demographic and intervention updates). Most other modules `update_results` near the end of `sim.step`. It's not as easy as calling update results because demographic modules have customized `update_results`, these will need to conform.

kaa2102:
This would be an good topic to discuss during a Starsim meeting. In seeking to align Births with Pregnancies/Deaths, I want to store data on who the mothers are and when they give birth. I was considering (1) creating a property sim.people.mother and sim.people.ti_mother or (2) using sim.results.

cliffckerr:
Closed by #504

PR #504, opened 2024-05-13
Demographic update_results normalization

daniel-klein:
### Description

* Normalizing `update_results` function for demographic modules, resolving #497.
* Also found and fixed a bug in the uids that are used to choose birth sex from pregnancy.


cliffckerr (approved):
Looks good!

PR #509, opened 2024-05-15
Prenatal and postnatal transmission

robynstuart:
### Description
Separates maternal transmission into prenatal and postnatal modules.

This is necessary because transmission probabilities vary. We could alternatively change the network betas, but this way allows us more flexibility to have disease-dependent differences - e.g. maternal transmission of syphilis only occurs during pregnancy, whereas for HIV it can also happen during breastfeeding.

### Checklist

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,6 +7,14 @@ What's new
 All notable changes to the codebase are documented in this file. Changes that may result in differences in model output, or are required in order to run an old parameter set with the current version, are flagged with the term "Regression information".
 
 
+Version 0.5.1 (2024-05-15)
+--------------------------
+- Separates maternal transmission into prenatal and postnatal modules
+- *GitHub info*: PR `509 <https://github.com/starsimhub/starsim/pull/509>`_
+
+
+
+
 Version 0.5.0 (2024-05-14)
 --------------------------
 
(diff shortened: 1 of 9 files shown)


daniel-klein:
What is the benefit of a separate network layers for prenatal and postnatal if all of the nodes and edges are the same as the current pregnancy network? Instead, wouldn't it be more straightforward for transmission, either via `rel_trans` of the mother, 'rel_sus' of the baby, or 'beta' for the edge, to vary as a function of the age of the baby? I suspect that transmission likely _should_ vary with age in any case through pre- and post-natal periods.

daniel-klein:
For the record, @robynstuart, @cliffckerr, and I discussed the following on Slack.

The transmission parameter, `beta`, associated with each edge applies to all diseases, so cannot be used to differentially transmit one disease and not another. The other two key parameters that control transmissibility are `rel_sus` and `rel_trans`, which are individual-level parameters. Therefore, we would not want to modify `rel_trans` of a mother during the pre-natal period because doing so would alter horizontal transmission. Modifying `rel_sus` for each baby and for each disease as a function of age would work.

This PR splits the pregnancy network into two: a prenatal network and a postnatal network. This additional resolution can facilitate disease-differential transmission during these periods, but it may still be necessary to modify `rel_sus` of the baby to e.g. adapt transmission by trimester.

PR #525, opened 2024-05-24
Ensure unborn agents cannot become pregnant.

daniel-klein:
### Description

Simple update to ensure unsure agents cannot become pregnant.

### Checklist

### DIFF
--- a/starsim/demographics.py
+++ b/starsim/demographics.py
@@ -345,6 +345,7 @@ class Pregnancy(Demographics):
         # Scale from rate to probability. Consider an exponential here.
         fertility_prob = fertility_rate * (self.pars.units * self.pars.rel_fertility * sim.pars.dt)
         fertility_prob[self.pregnant.uids] = 0 # Currently pregnant women cannot become pregnant again
+        fertility_prob[uids[self.sim.people.age[uids] <= 0]] = 0 # Unborn babies cannot become pregnant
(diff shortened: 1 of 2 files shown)


starsim/demographics.py
```diff
@@ -345,6 +345,7 @@ def make_fertility_prob_fn(self, sim, uids):
         # Scale from rate to probability. Consider an exponential here.
         fertility_prob = fertility_rate * (self.pars.units * self.pars.rel_fertility * sim.pars.dt)
         fertility_prob[self.pregnant.uids] = 0 # Currently pregnant women cannot become pregnant again
+        fertility_prob[uids[self.sim.people.age[uids] <= 0]] = 0 # Unborn babies cannot become pregnant
```
cliffckerr, line 348: This is obviously an improvement on the current implementation (yikes!), but in lieu of a full fecundability-by-age function, should we just have upper and lower age limits as a boxcar function? Having 1-year-olds and 99-year-olds get pregnant is also not ideal. We could potentially use [15,50] as default bounds?

daniel-klein: This is part of Pregnancy, for which users can specify either a fixed rate or age/year-specific fertility rates. Regardless, I would like to impose a hard constraint that unborn agents cannot become pregnant, just like already-pregnant agents. If users specify ASFR, this constraint might not have an effect, but is still worth including especially for the case in which the user provides a crude rate.

PR #526, opened 2024-05-28
Add option of sir_vaccine to be leaky or all_or_nothing

roberthinch:
### Description
Extend the sir_vaccine to support all_or_nothing (i.e. some vaccinated individuals are not susceptible whilst vaccine completely fails in others) as well as leaky (i.e. susceptibility in all vaccinated individuals reduced by the efficacy).

Added tests for the new all_or_nothing feature and the previous leaky behaviour.


cliffckerr (approved):
Looks good, thank you! Am just going to rename `all_or_nothing` to `leaky` (with flipped sign) since Starsim tries to use short names for things.

PR #528, opened 2024-05-30
Resolves ss.births not crn safe

kaa2102:
### Checklist


kaa2102:
The code has been refactored to make births stochastic (bernoulli) and linked to mothers. Also, the `sim.people.mother_uids` property is of type FloatArr (IndexArr generated indexing errors). 

On testing, the test_constant_pop() and test_nigeria() `asserts` fail as fewer births (constrained by available mothers) occur. Any suggestions on updating the test given the changes, e.g., changing the `death_rate`, the `np.isclose()` tolerance, etc?

cliffckerr:
Closing for now since tests fail due to changing the definition of fertility meaning population size is no longer matched, but can revisit branch later

Issue #507, opened 2024-05-14
Modulus-based random numbers induces unwanted correlations


daniel-klein:
We use r = (r1 + r2) % 1 for transmission, where r1 is a random number from the first agent and r2 is a random number from the second agent. r is indeed uniformly distributed on the interval from [0,1). However, when reusing the same per-agent random draws to test transmission on multiple network edges, we encounter correlations. This makes intuitive sense because a network can have order N^2 edges, but we're only drawing order N random numbers.

The correlation can be seen with as few as N=4 nodes:
```python
import numpy as np

N = 4
p = 0.5
R = 1000000

r = np.random.rand(4, R)
for method in ['indep', 'modulo']:
    if method=='modulo':
        p01 = (r[0,:] + r[1,:]) % 1
        p02 = (r[0,:] + r[2,:]) % 1
        p23 = (r[2,:] + r[3,:]) % 1
        p13 = (r[1,:] + r[3,:]) % 1
    else:
        p01 = r[0,:]
        p02 = r[1,:]
        p23 = r[2,:]
        p13 = r[3,:]

    alone = np.count_nonzero(p13<p) / R
    combined = np.count_nonzero((p13<p) & (p01<p) & (p02<p) & (p23<p)) / np.count_nonzero((p01<p) & (p02<p) & (p23<p))
    print(method, alone, combined)
```

Output shows the bias:
```bash
indep 0.50018 0.5008594843094144
modulo 0.49979 0.668084936217859
```

daniel-klein:
I propose that we restore the previous node-based acquisition method.

cliffckerr:
I think this is a tricky issue. There are three things I'm still unclear on:
1. Whether this would be an issue in practice for large networks where `p_connection << 1`;
2. Whether the node-based acquisition method actually solves the issue.
3. How this issue relates to constructing CRN-safe networks in the first place.

### 1. Does it matter in practice?
The induced correlations between pairs of agents are proportional to 1/√n. With 1000 agents, they're on the order of 1-3%:
![image](https://github.com/starsimhub/starsim/assets/3239256/0b51aa3e-1ab1-4c5b-9102-1d2f4d1d5d9d)
```py
import numpy as np
import sciris as sc
import pylab as pl

n = 1000

r1 = np.random.rand(n,n)
v2a = np.random.rand(n)
v2b = np.random.rand(n).reshape((n,1))
r2a = np.tile(v2a, (n,1))
r2b = np.tile(v2b, (1,n))
[...]

PR #542, opened 2024-06-06
Revert transmission to "acquisition-based" from "modulo"

daniel-klein:
### Description
Reverting transmission. Performance looks VERY BAD in the benchmark, 10x more time, but a practical application in TBsim changed from 107s in main to 110s with this branch. Not sure what's going on with the benchmark (or TBsim), but otherwise happy with the changes. Because this part of the model is heavily used, we might consider number or even cython for further performance optimization.

### Checklist


tests/baseline.json
```diff
@@ -2,25 +2,25 @@
...
+    "sir_cum_infections": 13079.0,
+    "sis_n_susceptible": 4062.0,
+    "sis_n_infected": 6917.71568627451,
+    "sis_prevalence": 0.617184370003921,
+    "sis_new_infections": 207.57843137254903,
+    "sis_cum_infections": 21173.0,
+    "sis_rel_sus": 0.4716048846681062,
+    "n_alive": 10979.71568627451,
+    "new_deaths": 45.76470588235294,
+    "cum_deaths": 4630.0
```
cliffckerr, line 24: I was surprised that the results changed so little, based on my earlier toy implementation that seemed to imply a 10% difference in the number of infections ([here](https://github.com/starsimhub/starsim/issues/507#issuecomment-2143811645)). But then a subsequent toy investigation showed no difference ([here](https://github.com/starsimhub/starsim/issues/507#issuecomment-2156405926)).

I did a more in-depth exploration here:
https://github.com/starsimhub/starsim/issues/507#issuecomment-2156515941

TLDR, even with 10,000 replicates, there is no statistically significant difference in any result between this branch and current `main`.

cliffckerr (commented):
I just did a more thorough test which found that this branch and `main` seem to be statistically identical (even though a previous toy implementation I wrote showed that it should make a difference). I think this is the most confused I've ever been on a model implementation question. 🤣

daniel-klein:
I'm curious how much performance difference there really is in practice. On TBsim, the difference between modulo and acquisition-based (roulette) is negligible. But benchmark.json shows an order of magnitude difference. Did you get to the bottom of this difference with any of your follow-up testing?

cliffckerr:
Superseded by #546

PR #543, opened 2024-06-09
Revert transmission example

cliffckerr:
Illustrative PR -- ready for review, but do not merge!! -- showing a different way of implementing the "roulette" approach (that actually does away with the roulette part). The thinking here is that most of the time all we really need is to know the probability of infection per target agent, which isn't too expensive to calculate (but still needs to be numbafied, since I couldn't think of a way to do it without a for loop or `np.apply_along_axis()`). The source determination could be done if and only if the user requests the infection log as well, since I don't think it's used/stored otherwise (afaik?).

Not super pretty. Also, for some reason this implementation randomly segfaults sometimes 🤣 Still hoping we can get to the bottom of this properly!

### DIFF
--- a/starsim/disease.py
+++ b/starsim/disease.py
@@ -8,6 +8,7 @@ import starsim as ss
 import networkx as nx
 from operator import itemgetter
 import pandas as pd
+import numba as nb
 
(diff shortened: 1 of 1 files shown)


cliffckerr:
Superseded by #546

PR #546, opened 2024-06-11
Combining rands

### LINKED ISSUE #507: Modulus-based random numbers induces unwanted correlations

cliffckerr:
### Description
Implements @RomeshA's proposal for combining two arrays of random numbers. Now working and ready for review!

Closes #507 and supersedes #542 and #543.

### Checklist


RomeshA:
Ah I see - it looks alright on my Windows machine but not on my Mac. The difference is the integer types - if I change it to 

```
    a1 = (a*1e9).astype(np.int32)
    b1 = (b*1e9).astype(np.int32)
```

then I get

image

I guess this what's happening is that the `a1*b1` being fitted into a smaller int is reflecting a whole lot of very large numbers back into a smaller range, and thereby getting something approximating a uniform distribution over the range of permitted ints. So the requirement is probably that the multiplication (here by 1e9) is large enough for that to happen. If you take the example above and change it to 1e3 then it reverts to the behaviour originally seen with int64

RomeshA:
Thanks for updating @cliffckerr , actually one other change we should make, since this algorithm relies on the integer overflow we should probably use `uint32` rather than `int32` because C/C++ has defined overflow behavior for unsigned ints but not for signed ints

RomeshA:
Oh sorry and a related point would then be to replace

```
c = c-c.min()
c = c/c.max()
```

with

```
c = c.astype(float)/4294967295
```

since we know the range of possible values and can scale based on the possible range rather than the values that we happen to generate

cliffckerr:
Hmm we're close but not quite there yet. `np.random.rand()` is on the range `(0,1)`, whereas this transformation stretches it to `[0,1]` (and, oddly, seems to include exactly one 0.0 and exactly one 1.0 entry for reasons I do not understand):
```py
import numpy as np
import starsim as ss

np.random.seed(59685)
n = int(1e6)

a = np.random.rand(n)
b = np.random.rand(n)
c = ss.combine_rands(a, b)
arrs = list(dict(a=a, b=b, c=c).items())

print('Min:', {k:v.min() for k,v in arrs})
print('Max:', {k:v.max() for k,v in arrs})
print('n==0', {k:(v==0).sum() for k,v in arrs})
print('n==1', {k:(v==1).sum() for k,v in arrs})

# Min: {'a': 2.0834123509727576e-06, 'b': 3.7607038394504855e-07, 'c': 0.0}
# Max: {'a': 0.9999978108412808, 'b': 0.9999995045043658, 'c': 1.0}
# n==0 {'a': 0, 'b': 0, 'c': 1}
# n==1 {'a': 0, 'b': 0, 'c': 1}
```

RomeshA:
The exactly one 0.0 and exactly one 1.0 is because of that `-c.min()` and `+c.max()` noted above. We can guarantee not to include 1.0 by dividing by 4294967296 rather than 4294967295. Perhaps a bit trickier to guarantee 0 isn't included, I guess we could add 1/4294967296 or something (no point doing `c = (c+1).astype(float)` because we might just cause an overflow that produces more 0's)

RomeshA:
So maybe something like `c = (1+c.astype(float))/4294967297` - haven't fully thought through whether there's an issue with floating point precision there though - probably not with a float64?

cliffckerr:
I'm updating it to just use the min and max values from the `a` array ... should be fine??

RomeshA:
In my mind, we're generating uniform random numbers by approximately uniformly randomly setting the bits of the `uint32` so it would make the most sense to have the scaling done based on the range of values that a `uint32` can take, independent of any of the sampled values. One side effect of having it depend on `a` is that if a draw happens to be the maximum or minimum then it would shift all of the other sampled values, which wouldn't be desirable from a CRN perspective!

cliffckerr:
Ok I _think_ this version is good to go :laughing:

RomeshA (approved):

tests/test_plugins.py
```diff
@@ -50,7 +50,7 @@ def update(self):
 
 class Penicillin(ss.Intervention):
     """ Create a penicillin (BPG) intervention for treating syphilis """
-    def __init__(self, year=2020, prob=0.5):
+    def __init__(self, year=2020, prob=0.8):
```
cliffckerr, line 53: Test was failing otherwise, think it was just unlucky stochastic variability

daniel-klein:
Are you getting `RuntimeWarning`s from the overflows in the multiply and subtract? My terminal is flooded:
```bash
/Users/dklein/GIT/starsim/starsim/utils.py:314: RuntimeWarning: overflow encountered in scalar multiply
  c = np.bitwise_xor(a1*b1, a1-b1)
/Users/dklein/GIT/starsim/starsim/utils.py:314: RuntimeWarning: overflow encountered in scalar subtract
  c = np.bitwise_xor(a1*b1, a1-b1)
/Users/dklein/GIT/starsim/starsim/utils.py:314: RuntimeWarning: overflow encountered in scalar multiply
  c = np.bitwise_xor(a1*b1, a1-b1)
/Users/dklein/GIT/starsim/starsim/utils.py:314: RuntimeWarning: overflow encountered in scalar subtract
  c = np.bitwise_xor(a1*b1, a1-b1)
/Users/dklein/GIT/starsim/starsim/utils.py:314: RuntimeWarning: overflow encountered in scalar multiply
  c = np.bitwise_xor(a1*b1, a1-b1)
/Users/dklein/GIT/starsim/starsim/utils.py:314: RuntimeWarning: overflow encountered in scalar subtract
  c = np.bitwise_xor(a1*b1, a1-b1)
/Users/dklein/GIT/starsim/starsim/utils.py:314: RuntimeWarning: overflow encountered in scalar multiply
  c = np.bitwise_xor(a1*b1, a1-b1)
/Users/dklein/GIT/starsim/starsim/utils.py:314: RuntimeWarning: overflow encountered in scalar subtract
  c = np.bitwise_xor(a1*b1, a1-b1)
/Users/dklein/GIT/starsim/starsim/utils.py:314: RuntimeWarning: overflow encountered in scalar multiply
  c = np.bitwise_xor(a1*b1, a1-b1)
/Users/dklein/GIT/starsim/starsim/utils.py:314: RuntimeWarning: overflow encountered in scalar subtract
  c = np.bitwise_xor(a1*b1, a1-b1)
/Users/dklein/GIT/starsim/starsim/utils.py:314: RuntimeWarning: overflow encountered in scalar multiply
  c = np.bitwise_xor(a1*b1, a1-b1)
/Users/dklein/GIT/starsim/starsim/utils.py:314: RuntimeWarning: overflow encountered in scalar subtract
  c = np.bitwise_xor(a1*b1, a1-b1)
/Users/dklein/GIT/starsim/starsim/utils.py:314: RuntimeWarning: overflow encountered in scalar multiply
  c = np.bitwise_xor(a1*b1, a1-b1)
/Users/dklein/GIT/starsim/starsim/utils.py:314: RuntimeWarning: overflow encountered in scalar subtract
  c = np.bitwise_xor(a1*b1, a1-b1)
/Users/dklein/GIT/starsim/starsim/utils.py:314: RuntimeWarning: overflow encountered in scalar multiply
  c = np.bitwise_xor(a1*b1, a1-b1)
/Users/dklein/GIT/starsim/starsim/utils.py:314: RuntimeWarning: overflow encountered in scalar subtract
  c = np.bitwise_xor(a1*b1, a1-b1)
/Users/dklein/GIT/starsim/starsim/utils.py:314: RuntimeWarning: overflow encountered in scalar multiply
  c = np.bitwise_xor(a1*b1, a1-b1)
...
```

daniel-klein:
Seems much better than Modulo! Hard to know without a statistical test if it is producing results consistent with Random.

image

from
```python
import networkx as nx
import numpy as np
import matplotlib.pyplot as plt
import pandas as pd
import seaborn as sns
from pandas.util import hash_pandas_object
import hashlib
import sciris as sc

from starsim.utils import combine_rands

import warnings
warnings.filterwarnings("ignore", "overflow encountered in scalar subtract")
warnings.filterwarnings("ignore", "overflow encountered in scalar multiply")
 
n = 8 # Number of nodes
n_sources = 2 # Number of sources (seed infections)
reps = 1_000_000
edge_prob = 0.7 # Edge probability
trans_prob = 0.4
seed = 1

def hash(df):
    #return int(hashlib.sha256(hash_pandas_object(df, index=True).values).hexdigest(), 16)
    return hashlib.sha256(hash_pandas_object(df, index=True).values).hexdigest()[:6]

def random(G):
    src = nx.get_node_attributes(G, 'infected')
    infected = []
    tx = []

    # Randomize in case a node is infected multiple times
    el = list(G.edges())
    np.random.shuffle(el)

    for (n1,n2) in el:
        # n1 --> n2
        if src[n1] and (not src[n2]) and (n2 not in infected) and (np.random.rand() < trans_prob):
            tx.append((n1, n2))
            infected.append(n2)
        # n2 --> n1
        if src[n2] and (not src[n1]) and (n1 not in infected) and (np.random.rand() < trans_prob):
            tx.append((n2, n1))
            infected.append(n1)
    return tx


def modulo(G):
    src = nx.get_node_attributes(G, 'infected')
    infected = []
    tx = []
    r1 = np.random.rand(n)
    r2 = np.random.rand(n)
    el = list(G.edges())
    np.random.shuffle(el)
    for (n1,n2) in el:
        # n1 --> n2
        if (src[n1]) and (not src[n2]) and (n2 not in infected) and (((r1[n1]+r2[n2])%1) < trans_prob):
            tx.append((n1, n2))
            infected.append(n2)
        # n2 --> n1
        if (src[n2]) and (not src[n1]) and (n1 not in infected) and (((r1[n2]+r2[n1])%1) < trans_prob):
            tx.append((n2, n1))
            infected.append(n1)
    return tx


def combine_hash(G):
    src = nx.get_node_attributes(G, 'infected')
    infected = []
    tx = []
    r1 = np.random.rand(n)
    r2 = np.random.rand(n)
    el = list(G.edges())
    np.random.shuffle(el)
    for (n1,n2) in el:
        # n1 --> n2
        r = combine_rands(r1[n1], r2[n2])
        if (src[n1]) and (not src[n2]) and (n2 not in infected) and (r < trans_prob):
            tx.append((n1, n2))
            infected.append(n2)
        # n2 --> n1
        r = combine_rands(r1[n2], r2[n1])
        if (src[n2]) and (not src[n1]) and (n1 not in infected) and (r < trans_prob):
            tx.append((n2, n1))
            infected.append(n1)
    return tx

def roulette(G):
    src = nx.get_node_attributes(G, 'infected')
    tx = []
    for n2 in G.nodes():
        # All nodes --> n2
        if not src[n2]:
            srcs = [n for n in G.neighbors(n2) if src[n]] #[src[n] for n in G.neighbors(n2)] # Count infected neighbors
            cnt = len(srcs)
            if np.random.rand() < 1-(1-trans_prob)**cnt:
                n1 = np.random.choice(srcs)
                tx.append((n1, n2))
    return tx
 

def transmit(G, trans_fn):
    # Align transmissions from tx_in if provided
    txs = {}
    counts = {}
 
    for _ in np.arange(reps):
        txl = trans_fn(G)
        tx = pd.DataFrame(txl, columns=['src', 'dst']).sort_values(['src', 'dst']).reset_index(drop=True)
        
        h = hash(tx)

        if h not in txs:
            txs[h] = tx

        counts[h] = counts.get(h, 0) + 1

    df = pd.DataFrame(counts.values(), index=pd.Index(counts.keys(), name='Hash'), columns=['Counts'])
    return txs, df


# Build the graph
G = nx.random_graphs.erdos_renyi_graph(n=n, p=edge_prob, seed=seed)

# Seed infections
infected = {i:False for i in range(n)}
sources = np.random.choice(a=range(n), size=n_sources, replace=False)
for source in sources:
    infected[source] = True
nx.set_node_attributes(G, infected, 'infected')
 
# Do transmissions via each method in parallel
results = sc.parallelize(transmit, iterkwargs=[{'trans_fn':random}, {'trans_fn':modulo}, {'trans_fn':combine_hash}, {'trans_fn':roulette}], kwargs={'G':G}, die=True, serial=False)
tx, cnt = zip(*results)

df = pd.concat(cnt, axis=1) \
    .fillna(0) \
    .astype(int)
df.columns = ['Random', 'Modulo', 'CombineHash', 'Roulette']

# Manipulate results
df.reset_index(inplace=True)
dfm = df.melt(id_vars='Hash', var_name='Method', value_name='Count')

# Plot
g = sns.barplot(data=dfm, x='Hash', y='Count', hue='Method')
plt.xticks(rotation=90)
g.figure.tight_layout()

txc = tx[0].copy()
for i in range(1, len(tx)):
    txc.update(tx[i])
for h, v in txc.items():
    print(f'\nUnique transmission tree #{h}')
    print(v)

plt.figure()
pos = nx.spring_layout(G, seed=3113794652)  # positions for all nodes
ec = nx.draw_networkx_edges(G, pos, alpha=0.2)
colors = ['red' if infected else 'blue' for infected in nx.get_node_attributes(G, 'infected').values()]
nc = nx.draw_networkx_nodes(G, pos, nodelist=G.nodes(), node_color=colors, node_size=100)
nx.draw_networkx_edges(G, pos, width=1.0, alpha=0.5)
nx.draw_networkx_labels(G, pos)

plt.show()
```

daniel-klein:
I'll add that code as a dev_test.

daniel-klein:
So other than the large number of warning message, this looks great and also means we can implement CRN-safe random networks (using a similar approach).

daniel-klein (approved):
Looks good other than the warning messages, which can be suppressed.

daniel-klein:
According to [numpy documentation](https://numpy.org/doc/stable/reference/random/generated/numpy.random.rand.html), `rand()` returns a value in range [0,1). Therefore, I will adjust the code slightly to ensure that 0 is possible.

cliffckerr:
@daniel-klein I think this is risky ... in practice you would need to draw quadrillions of values to ever encounter a zero:
>>> print(np.random.rand(int(1e9)).min())
2.3877175792108574e-09
```
In contrast, since this method is discretized by `uint32` integers, the chance is, well, 2^32 times higher, which is a big number!

I wonder if there is a way to keep the core principles of this method while avoiding (a) the overflow, (b) this discretization? @RomeshA, thoughts?

**Comment by daniel-klein** [comment]
Wait, if we want uniformly distributed uint32 numbers instead of continuous uniform, why not generate those in the first place? So this would avoid the need for the following two lines:
```python
a1 = (a*1e9).astype(np.uint32)
b1 = (b*1e9).astype(np.uint32)
```
as inputs `a` and `b` would come from `random.randint`.

daniel-klein:
Okay, it seems to work and now doesn't have overflow/underflow error as the conversion from rand to int happens upstream.

starsim/disease.py
```diff
@@ -150,8 +150,8 @@ def __init__(self, *args, **kwargs):
         )
 
         # Define random number generators for make_new_cases
-        self.rng_target = ss.random(name='target')
-        self.rng_source = ss.random(name='source')
+        self.rng_target = ss.randint(low=np.iinfo(np.int64).min, high=np.iinfo(np.int64).max, dtype=np.int64, name='target')
+        self.rng_source = ss.randint(low=np.iinfo(np.int64).min, high=np.iinfo(np.int64).max, dtype=np.int64, name='source')
```
cliffckerr, line 154: We'll need to make this safe for switching Starsim dtypes in the settings, but otherwise this looks great!

cliffckerr (commented):
Need to remove the hard-coding to int64 but otherwise looks good!

starsim/disease.py
daniel-klein: Do we though? Not sure why a default dtype should matter as this feels independent. What's the potential issue?

daniel-klein: Also, looking back at this code with fresh eyes, I don't remember why I picked `int64` over `uint64` here. Probably doesn't make a difference other than possibly some runtimewarnings on some OS? I switched over to numpy's random integer implementation precisely because it allows the user to specify the data type, scipy does not. Do we have a preference?

RomeshA: I am a little puzzled as to why there is no overflow now, I would have thought that `a*b` could still overflow for large values of `a` and `b`? My thinking in picking `uint` was then to make sure that the `a*b` overflow was a defined operation

daniel-klein: I think I now understand the warning messages. When passing numpy arrays to `combine_rands`, all is well. But if slicing to individual (`np.float32`) values and then calling the function, the warning is triggered.

daniel-klein: Instead of `(a*1e9).astype(np.uint32)`, I was thinking here that we could directly generate uniformly distributed random integers, initially `uint32`. When @cliffckerr described the issues with overrepresentation of 0 and 1 due to quantization, I realized we could directly generate 64-bit random integers and then use the bitwise xor. Any opinions on this and if we should generate signed or unsigned integers?

Issue #548, opened 2024-06-14
RandomNet and degree


daniel-klein:
RandomNet halves the desired degree for each now, and casts to integer. But this is challenging if the desired degree is 1, which will get changed to a zero, meaning no chance of infection even though the desired degree was 1.

Recommend:
```python
number_of_contacts = sc.randround(number_of_contacts / 2).astype(ss_int_)  # One-way contacts
```

This way, agents with a desired degree of 1 will get 0 or 2, but that's still better than always 0.

cliffckerr:
I ran into this issue as well. I do wonder if we should rethink the definition of number of contacts, because it is confusing that odd numbers are basically not supported.

daniel-klein:
I was using a Poisson distribution with a low mean, so most values were zero or one --- and was getting basically ZERO transmission!

cliffckerr:
Fixed by #550

Issue #547, opened 2024-06-14
RandomNet doesn't know about pregnancy


daniel-klein:
With RandomNet, edges are formed to unborn agents.

cliffckerr:
Fixed by #550

PR #550, opened 2024-06-14
Fix RandomNet and add get_analyzer

daniel-klein:
### Description

- Adjusting ``RandomNet`` to avoid connections to unborn agents and use random rounding for half edges. Addresses #547 and #548.
- Adds ``get_analyzers`` and ``get_analyzer``

### Checklist


cliffckerr (approved):
Looks good!

PR #554, opened 2024-06-18
Performance tweaks

RomeshA:
### Description

This PR refactors how data is pre-processed for births/pregnancy/death rates, overall this gives about a 10% decrease in run time for the STIsim HIV model

Also a minor change to `uids` logical operations to simplify mixing working with `uids` and `BoolArr`s where `BoolArr.uids` is automatically called when doing set operations on `uids` with a `BoolArr`

### Checklist


cliffckerr (commented):
Looks good!

Issue #510, opened 2024-05-16
Add label to results


cliffckerr:
e.g. `Cumulative HIV infections`, so results are easier to plot in a self-contained way

cliffckerr:
Closed by #560

PR #560, opened 2024-06-18
Add label to results Class

devclinton:
### Description
- Added label to the Result class (#510) 
- Update existing definitions of States to have accompanying labels 
- Update plot function of sim to leverage the new results labels

I am not sure if I should split the updates to existing modules to another ticket? Also, I am not 100% confident in my labels, so extra attention would be greatly appreciated. 

TODO: Update tests to include results

### DIFF
--- a/starsim/demographics.py
+++ b/starsim/demographics.py
@@ -72,9 +72,9 @@ class Births(Demographics):
     def init_results(self):
         npts = self.sim.npts
         self.results += [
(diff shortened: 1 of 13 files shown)


cliffckerr (commented):
Looks good -- I don't know if we want to add labels to all states, but this is definitely a big improvement, and they shouldn't hurt!

PR #570, opened 2024-06-27
Improving pregnancy configuration and fixing test_randomness

daniel-klein:
Fixing demographics so that users can override the fertility rate parameter even if using a subclass. Now `self.pars.p_fertility` in `init_pre` rather than doing everything in `__init__`.

Also changing `uint64` to signed `int64` in test_randomness.

### Checklist


daniel-klein:
Tests not passing due to crude handling of scalar and DataFrame inputs to demographics. Might need a more significant reworking.

cliffckerr (approved):
Looks good, just changed `isinstance(a, x) or isinstance(a, y)` to `isinstance(a, (x,y))`

PR #573, opened 2024-06-30
Logging off by default

### LINKED ISSUE #569: Disable InfectionLog by default
Requires significant time (~20% of sim total), and typically not used. Should be optional. There is already a `log=True` arg to `Disease.__init__()`, but unfortunately it's hard to get this parameter passed with the way kwargs are currently processed. Need to figure out a better solution.

RomeshA:
### Description

This PR changes `log` from an argument of `Disease` to being part of the pars, so that it can be passed more easily through inheritance and handled alongside any other module settings. The `InfectionLog` is still created so that it can be used by custom code even if standard logging is not enabled (this is what STIsim does) but entries are only added to it by Starsim if `log=True`

### Checklist

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,6 +7,14 @@ What's new
 All notable changes to the codebase are documented in this file. Changes that may result in differences in model output, or are required in order to run an old parameter set with the current version, are flagged with the term "Regression information".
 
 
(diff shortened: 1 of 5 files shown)


cliffckerr (approved):
Excellent!

PR #575, opened 2024-07-01
New CRN networks

daniel-klein:
### Description

Adding two new CRN-safe networks. The first is an Erdos-Renyi network using the same logic as disease transmission. The second is a disk graph between agents that randomly bounce around in a unit cube.

Simple tests implemented.


cliffckerr (commented):
From a quick glance looks good! Do we need to keep both `RandomNet` and `ErdosRenyiNet`?

RomeshA:
I guess an advantage of `RandomNet` would be if you didn't want the binomial distribution of node degree in the `ErdosRenyiNet` (e.g., if you wanted lognormal, or to base it on data etc.)?

starsim/network.py
```diff
@@ -524,6 +524,141 @@ def add_pairs(self):
...
+    """
+    Disk graph in which edges are made between agents located within a user-defined radius.
+
+    Interactions take place within a square with edge length of 1. Agents are
+    initialized to have a random position and orientation within this square. On
+    each time step, agents advance v*dt in the direction they are pointed. When
+    encountering a wall, agents are reflected.
+
+    Edged are formed between two agents if they are within r distance of each other.
+    """
```
cliffckerr, line 589: I think this is cool but am struggling a little with a use case for it. I think the most useful change would be to be able to have connections as a distance-dependent probability rather than (or in addition to) a boxcar. Since you're already calculating the full distance matrix, this should be relatively easy to do?

The second question is about the constant velocity with edge reflection. I think (as you mentioned in the meeting) that this could lead to strange statistical artifacts in terms of who encounters whom over time. Why not have it be a random walk instead?

starsim/network.py
```diff
@@ -524,6 +524,141 @@ def add_pairs(self):
...
+        """ Initialize """
+        super().__init__(key_dict=key_dict)
+        self.default_pars(
+            r = 0.1, # Radius
+            v = 0.05, # Velocity
+        )
+        self.update_pars(pars, **kwargs)
+        self.add_states(
+            ss.FloatArr('px', default=ss.random(), label='X position'),
+            ss.FloatArr('py', default=ss.random(), label='Y position'),
```
cliffckerr, line 601: We often use `p` for 'probability', so maybe just `x` and `y` would be clearer? (No issue with name collisions since states when centralized also inherit their module names.)

starsim/network.py
```diff
@@ -524,6 +524,141 @@ def add_pairs(self):
...
+    def init_post(self):
+        self.add_pairs()
+        return
+
+    def update(self):
+
+        # Motion step
+        vdt = self.pars.v * self.sim.dt
+        self.px[:] = self.px + vdt * np.cos(self.theta)
+        self.py[:] = self.py + vdt * np.sin(self.theta)
```
cliffckerr, line 615: You should be able to use `+=` notation here, which would simplify a lot of the below code

starsim/network.py
```diff
@@ -524,6 +524,141 @@ def add_pairs(self):
...
+
+        self.theta[:] = np.mod(self.theta, 2*np.pi)
+
+        self.add_pairs()
+        return
+
+    def add_pairs(self):
+        """ Generate contacts """
+        people = self.sim.people
+        born_uids = (people.alive & (people.age > 0)).uids
```
cliffckerr, line 647: As above (can be just `(people.age > 0).uids`)

starsim/network.py
```diff
@@ -524,6 +524,141 @@ def add_pairs(self):
...
+
+        pos = np.vstack([self.px, self.py]).T
```
cliffckerr, line 649: Slightly longer but more transparent to use `np.column_stack([self.px, self.py])`

tests/test_networks.py
```diff
@@ -6,8 +6,9 @@
 import sciris as sc
 import numpy as np
 import starsim as ss
+import scipy.stats as sps
 
-sc.options(interactive=False) # Assume not running interactively
+sc.options(interactive=True) # Assume not running interactively
```
cliffckerr, line 11: This should be set to `False` here, and then `sc.options(interactive=True)` can be in the `main` block (no need to slow down automated `pytest` tests with plotting).

tests/test_networks.py
```diff
@@ -61,6 +62,79 @@ def test_random():
     o = sc.objdict(nw1=nw1, nw2=nw2, nw3=nw3)
     return o
 
+def test_erdosrenyi():
+    sc.heading('Testing Erdos-Renyi network')
+
+    def test_ER(n, p, nw):
```
cliffckerr, line 68: A quick docstring explaining how this is testing the network would be good

tests/test_networks.py
```diff
@@ -61,6 +62,79 @@ def test_random():
...
+        ax.scatter(nw1.px, nw1.py, s=50, c=colors)
+        for i in range(s1.pars.n_years):
+            ax.plot([0,1,1,0,0], [0,0,1,1,0], 'k-', lw=1)
+            ax.quiver(nw1.px, nw1.py, vdt * np.cos(nw1.theta), vdt * np.sin(nw1.theta), color=colors)
+            ax.set_aspect('equal', adjustable='box') #ax.set_xlim([0,1]); ax.set_ylim([0,1])
+            s1.step()
+
+    # Simulate SIR on a DiskNet
+    nw2 = ss.DiskNet(r=0.15, v=0.05)
+    s2 = ss.Sim(n_agents=small, networks=nw2, diseases=[ss.SIR()], copy_inputs=False).initialize() # This initializes the network
```
cliffckerr, line 129: This can just be `ss.Sim(n_agents=small, networks=nw2, diseases='sir').initialize()`

tests/test_networks.py
```diff
@@ -61,6 +62,79 @@ def test_random():
...
+    return o
+
+def test_disk():
+    sc.heading('Testing Disk network')
+
+    # Visualize the path of agents
+    nw1 = ss.DiskNet()
+    s1 = ss.Sim(n_agents=5, n_years=50, networks=nw1, copy_inputs=False).initialize() # This initializes the network
+
+    if sc.options.interactive:
```
cliffckerr, line 110: In other tests we usually use `if do_plot:` ... I guess this works too though!

cliffckerr (commented):
More thorough review this time! Very nice to have these two new networks. None of the comments are breaking/urgent, so I think the two options are we can either merge this as-is and make issues as needed, or try to address them prior to merge. I don't have a strong preference either way.

starsim/network.py
```diff
@@ -524,6 +524,141 @@ def add_pairs(self):
...
 
+class ErdosRenyiNet(DynamicNetwork):
+    """ Random connectivity between agents """
+
+    def __init__(self, pars=None, key_dict=None, **kwargs):
+        """ Initialize """
+        super().__init__(key_dict=key_dict)
+        self.default_pars(
+            p = 0.1, # Probability of each edge
+            dur = 0, # Duration of zero ensures that new random edges are formed on each time step
```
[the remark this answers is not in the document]
daniel-klein: YES! New edges will be formed on every time step regardless if the old edges are removed or not. I wasn't sure if we should disallow `dur` (inherit from `Network`?), but ended up leaving it in. USER BEWARE!

starsim/network.py
```diff
@@ -524,6 +524,141 @@ def add_pairs(self):
...
+        )
+        self.update_pars(pars, **kwargs)
+        self.randint = ss.randint(low=np.iinfo('int64').min, high=np.iinfo('int64').max, dtype=np.int64) # Used to draw a random number for each agent as part of creating edges
```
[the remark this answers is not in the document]
daniel-klein: I don't know, but have done extensive testing using the full signed 64-bit range so would strongly prefer to keep as is. It can't be better if not using the negative integers.

starsim/network.py
```diff
@@ -524,6 +524,141 @@ def add_pairs(self):
         return
 
 
+class ErdosRenyiNet(DynamicNetwork):
```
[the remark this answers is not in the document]
daniel-klein: Agree with accessibility, and think we can solve it with the docstring.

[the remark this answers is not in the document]
daniel-klein: I hope to do some performance testing soon, but it wouldn't surprise me if this ERN is speed-competitive with `RandomNet`.

starsim/network.py
```diff
@@ -524,6 +524,141 @@ def add_pairs(self):
...
+
+    def update(self):
+        self.end_pairs()
+        self.add_pairs()
+        return
+
+    def add_pairs(self):
+        """ Generate contacts """
+        people = self.sim.people
+        born_uids = (people.alive & (people.age > 0)).uids
```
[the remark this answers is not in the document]
daniel-klein: Ohhh, that's an interesting idea. But I'd worry about vertical transmission to not-alive (uborn) agents.

starsim/network.py
```diff
@@ -524,6 +524,141 @@ def add_pairs(self):
...
+        edge = r <= self.pars.p
+
+        p1 = idx1[edge]
+        p2 = idx2[edge]
+        beta = np.ones(len(p1), dtype=ss_float_)
+
+        if isinstance(self.pars.dur, ss.Dist):
+            dur = self.pars.dur.rvs(p1)
+        else:
+            dur = np.full(len(p1), self.pars.dur)
```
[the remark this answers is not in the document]
daniel-klein: I like the scalar idea as it unifies the architecture. The `np.full` probably isn't even needed here...

starsim/network.py
daniel-klein: Neat idea! Yes, need not be a strict disk graph and we could easily allow the user to provide a function that maps distance to edge probability. It'd slow things down ever so slightly, but not a big deal.

daniel-klein: OH, I also LOVE the random walk idea! Perhaps as an option? Constant velocity or random walk? It should just amount to an `if` statement in `update`. Or subclass overriding `update`?

tests/test_networks.py
daniel-klein: Oops!

daniel-klein:
Thanks for the feedback @cliffckerr and @RomeshA. I'll push a few changes momentarily.

starsim/network.py
daniel-klein: With `+=`, I'm getting the ufunc error:
```python
RecursionError: maximum recursion depth exceeded while calling a Python object
```

tests/test_networks.py
daniel-klein: Will do.

PR #586, opened 2024-07-10
Interpolate asfr

alina-muellenmeister:
- Interpolate ASFR data to 1-year increments in `standardize_fertility_data()`

- Update initial `new_rates` in line 320 to shallow copy of the data to avoid re-assigning properties to the source object
- Remove `new_rates` assignment in 335 as `np.divide` stores results to `new_rates` automatically


RomeshA (approved):
Looks good to me, @cliffckerr this one should fix @robynstuart's bug with STIsim so would be helpful to merge this in soon

Issue #568, opened 2024-06-27
Improve transmission performance


cliffckerr:
PR #546 appears to slow down the sim by nearly 30% -- the benchmark takes 1.3 s to run instead of 1.0 s. Need to fix. Tried `njit` on `ss.combine_rands()`, doesn't help.

Compare SHA 4689698 (`combine-rands`, slow) with 43249e3 (`main`, fast).

cliffckerr:
Script:
```py
import sciris as sc
import starsim as ss

kw = dict(n_agents=10e3, start=2000, end=2100, diseases='sis', plot=False)

prof = 1

if prof:
    context = sc.cprofile(sort='selfpct', mintime=1e-1)
else:
    context = sc.timer()

with context:
    sim = ss.demo(**kw)
```

Slow:
```
                          func    cumpct  selfpct  cumtime    selftime  calls                                        path
0                     add_edge   14.6439  14.3256   0.1165  1.1399e-01  51710                         multidigraph.py:416
1                set_prognoses   20.3468   3.1097   0.1619  2.4744e-02    102                               disease.py:94
2               make_new_cases   49.8485   2.4438   0.3967  1.9446e-02    101                              disease.py:247
3                       append   16.5239   1.8800   0.1315  1.4959e-02  51710                              disease.py:377
4                          rvs   17.2986   1.8735   0.1376  1.4908e-02    611                        distributions.py:520
5             _handle_fromlist   13.5303   0.3817   0.1077  3.0370e-03   4897          <frozen importlib._bootstrap>:1207
6   load_additional_registries   15.4045   0.2278   0.1226  1.8130e-03    248                                   cpu.py:60
7                         step   73.3433   0.1977   0.5836  1.5733e-03    101                                  sim.py:167
8                set_prognoses   23.9644   0.1740   0.1907  1.3842e-03    102                                  sir.py:128
9                   _set_cases   24.0492   0.1045   0.1914  8.3182e-04    101                              disease.py:305
10                         run  100.5591   0.0824   0.8002  6.5575e-04      1                                  sim.py:234
11              _find_and_load   13.0682   0.0316   0.1040  2.5143e-04     88          <frozen importlib._bootstrap>:1165
12                      update   13.2567   0.0308   0.1055  2.4541e-04    101                              network.py:499
13     _find_and_load_unlocked   13.0037   0.0307   0.1035  2.4405e-04     76          <frozen importlib._bootstrap>:1120
[...]

PR #590, opened 2024-07-11
Raw bit generator sampling

RomeshA:
### Description

The issue observed in #568 appears to be because `rng.random()` (to get a random float) is faster than `rng.integers()`. It's been suggested that this is because `random()` is a pure C call while `integers()` has some Python logic as well. In the test case noted by @cliffckerr the simple model runtime is dominated by the transmission portion of the calculation and this difference is big enough to provide a noticeable difference in runtime.

@daniel-klein had suggested a while ago that maybe we could just use the numpy bit generator directly to get the random integers we need. I've implemented this in a new `Dist`, `rand_uint64`,  that calls `bitgen.random_raw` to get the `uint64` array of random integers that is then passed to `combine_rands`. We're already asking users who want to use `combine_rands` to create `Dist` instances with integers even though the final random number is going to be a float, so they have to view the setup as an integrated package (with integer `Dists` and `combine_rands`) so I don't think it's too onerous (and in fact, maybe a bit simpler/clearer) to just say `rand_uint64` provides inputs for use with `combine_rands` (could potentially give it a better name).

Performance is a little bit tricky because the current `main` branch has had other optimizations in the meantime. Running @cliffckerr 's test script from #568 gives

- Old `main` (`43249e3`): 733 ms ± 3.6 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
- `combine_rands` (`4689698`): **10% slower** (810 ms ± 6.61 ms per loop (mean ± std. dev. of 7 runs, 1 loop each))
- `main` (`4e0bcfc`): **12% faster** (641 ms ± 15.6 ms per loop (mean ± std. dev. of 7 runs, 1 loop each))
- `test_raw_rng`: **20% faster** (582 ms ± 4.67 ms per loop (mean ± std. dev. of 7 runs, 1 loop each))

So overall `test_raw_rng` has higher performance than the old `main` branch where `combine_rands` had worse performance. The impact of this issue depends on what proportion of time in a model is spent calculating transmission, so for a complex model like STIsim, it makes almost no difference to the runtime because so much time is spent on other parts of the model). 

Not sure about whether it's worth the added complexity of a separate integer generator just for this purpose, but on the other hand, the performance _is_ better and it solves #568, so opening this PR for comment/review

### Checklist


daniel-klein:
Thanks @RomeshA, this looks amazing. I saw the `random_raw` as well, but at that point was convinced that we needed signed integers. In response to your comment on the paper, I've done a bit more archeology and no longer see an advantage either way.

daniel-klein (approved):
Looks good! I just added a fix to enable `_centralized`, which I'm using for the CRN paper. Also just running a statistical test (from the paper) to make sure all is well, but assuming it will be fine.

daniel-klein:
Ut oh! Just ran a small test and it seems to produce highly biased results. Need to do some more investigation, but please do not merge yet!

image

daniel-klein (changes requested):
Trying to remove my "approval" due to the observed bias.

RomeshA:
@daniel-klein Weird! It looked fine with Cliff's 2D histogram tests, and a reduced run of `explore_tx_correlation.py` also seemed OK - what was the script that produced those biased results?

daniel-klein:
@RomeshA https://github.com/starsimhub/crn_paper/blob/main/explore_erdosrenyinet.py

daniel-klein:
I'm wondering if it's related to this warning in the [random_raw](https://numpy.org/doc/stable/reference/random/bit_generators/generated/numpy.random.BitGenerator.random_raw.html) documentation:
```
This method directly exposes the raw underlying pseudo-random number generator. All values are returned as unsigned 64-bit values irrespective of the number of bits produced by the PRNG.
```

For the failing test, I'm using the default Mersenne Twister.

daniel-klein:
Ah yes, that's probably it. From the [Mersenne Twister (MT19937)](https://numpy.org/doc/stable/reference/random/bit_generators/mt19937.html) docs:
```
The [MT19937](https://numpy.org/doc/stable/reference/random/bit_generators/mt19937.html#numpy.random.MT19937) state vector consists of a 624-element array of 32-bit unsigned integers plus a single integer value between 0 and 624 that indexes the current position within the main array.
```

daniel-klein:
Let me try with a different Generator

daniel-klein:
"The [PCG64DXSM](https://numpy.org/doc/stable/reference/random/bit_generators/pcg64dxsm.html#numpy.random.PCG64DXSM) state vector consists of 2 unsigned 128-bit values, which are represented externally as Python ints."

Not sure how well this will work, might just be inefficient to return as `uint64`, but running test now.

daniel-klein:
Yay, all good with PCG64DXSM. I can fix `_centralized` with something like the following
```python
np.random.mtrand._rand.randint(low=0, high=np.iinfo(np.uint64).max, dtype=np.uint64, size=n)
```

daniel-klein (approved):
So it depends on the Generator, with PCG-64 DXSM (our default) being okay but Mersenne Twister (MT19937) no good. Have created a workaround for `_centralized`.

RomeshA:
Great, thanks for looking into that @daniel-klein !

PR #596, opened 2024-08-20
Net chunking

daniel-klein:
### Description

Adding chunking to ErdosRenyiNet. Chunking trades memory for CPU, allowing the algorithm to run on larger networks, but at the cost of additional wall clock time.

NOTE: Redundancy in the chunking algorithm between ErdosRenyiNet and DiskNet that should be improved.

### Checklist


daniel-klein:
Note also includes changes from #595

cliffckerr:
I would prefer the O(N) algorithm if that's acceptable (it's not strictly identical to an E-R network, but should be close enough for all practical purposes).

cliffckerr (commented):
I would prefer the O(N) algorithm if that's acceptable (it's not strictly identical to an E-R network, but should be close enough for all practical purposes).

cliffckerr:
Going to close this for now, but will leave the branch and we can revisit post-Symposium.

Issue #609, opened 2024-09-10
Consider EnumArr


cliffckerr:
Should we allow users to specify a single `Arr` with multiple discrete states? For example, do we have three separate S, I, R states, or a single state that can take on three values? What are the pros and cons of each? Is the answer "it depends" to the point that we should allow users to choose either approach? This would be a major refactor, but potentially worth it.

PR #610, opened 2024-09-10
timestep enum

menriquez-IDM:
### Description
A suggestion for starsim, it could be a useful feature to have for the users. Why I put it in a folder, because the idea is to have the same level of information, samples etc. for other enums which may come along. 
Feedback and other considerations are welcome. 

### Checklist

Not sure if it's worth to have a new version.

### DIFF
--- a/starsim/__init__.py
+++ b/starsim/__init__.py
@@ -18,6 +18,7 @@ from .sim           import *
 from .run           import *
 from .calibration   import *
 from .samples       import *
+from .enums.timestep  import TimeStep
 
 # Assign the root folder
 import sciris as sc
--- a/starsim/enums/__init__.py
+++ b/starsim/enums/__init__.py
@@ -1,0 +1,2 @@
+from .timestep         import *
+
--- a/starsim/enums/timestep.py
+++ b/starsim/enums/timestep.py
@@ -20,6 +20,26 @@ class TimeStep(Enum):
         __repr__() -> str: Returns a string representation of the enum member, which is its name.
         __float__() -> float: Returns the value of the enum member as a float.
         __str__() -> str: Returns a string representation of the value of the enum member.
+        
+    Examples:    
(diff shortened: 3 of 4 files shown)


menriquez-IDM:
@JSchripsema-IDM  this is a case where the preview documents Github action actually help :)  ... or is there a way to see why it failed?

JSchripsema-IDM:
@menriquez-IDM clicking "Details" should show you the build logs on RTD:

Configuration error:
There is a programmable error in your configuration file:

  File "/home/docs/checkouts/readthedocs.org/user_builds/institute-for-disease-modeling-starsim/envs/610/lib/python3.11/site-packages/sphinx/config.py", line 529, in eval_config_file
    exec(code, namespace)  # NoQA: S102
    ^^^^^^^^^^^^^^^^^^^^^
  File "/home/docs/checkouts/readthedocs.org/user_builds/institute-for-disease-modeling-starsim/checkouts/610/docs/conf.py", line 8, in <module>
    import starsim as ss
  File "/home/docs/checkouts/readthedocs.org/user_builds/institute-for-disease-modeling-starsim/envs/610/lib/python3.11/site-packages/starsim/__init__.py", line 21, in <module>
    from .enums.timestep  import TimeStep
ModuleNotFoundError: No module named 'starsim.enums'

menriquez-IDM:
I tried that however I get a 404, am I missing something? 
In any case, thanks for investigating for me.

cliffckerr:
I think the issue is with the subfolder. You could try adding an `__init__.py` file to the subfolder, or just move the file to the main folder.

However, the time refactor will mean that a year is no longer fixed as the base time unit, so we shouldn't need a enum for this. We have discussed having an `EnumArr` (https://github.com/starsimhub/starsim/issues/609), but we'll revisit that after v2.0.

Issue #601, opened 2024-09-09
Rethink connectors/interventions/analyzers/plugins


cliffckerr:
Replace connectors with a `Plugin` class that is also the parent class of interventions and analyzers (making an issue but actually already mostly done on one of my branches).

cliffckerr:
Closed by #612

Issue #326, opened 2024-02-27
Rethink time


cliffckerr:
We need to solve how we handle time. I was thinking is that each module could have a time unit defined, which at least initially we could restrict to 'day' or 'year'. Then all parameter values defined in that module would be relative to that module's time unit. (If `unit='day'`, then start and end would be dates; if `unit='year'`, they would be ints/floats, although dates would also work. I was also thinking of a 'none' option, which would just use the timesteps directly.)

It would be nice if we had very simple `ss.dur` and `ss.rate` classes, which are time-unit aware. This would save the user the trouble of remembering to divide/multiply by the correct `dt` everywhere. The user wouldn't have to see these; they'd be implemented in the default parameters and merged with user-supplied parameters. So the default for SIR would be `pars.dur_inf = ss.dur(6)`, but the user could just supply `diseases=dict(type='sir', dur_inf=6)` and it would get converted a dur based on the default (this is related to the sim API work).

The part I really struggled with was how to do the actual integration in `sim.step()`. Should the sim timestep be the smallest of the module timesteps, such that on each sim timestep, each module either runs or is `pass`? Or should the sim timestep equal the results timestep, which may or may not correspond to any individual module timestep? And how do we nicely handle results when everything is happening at different times? My best idea was something like this:
```py
def sim.step():
  sim.now += sim.dt
  for module in modules:
    module.step(sim)

def module.step(sim):
  while self.now < sim.now:
    self.now += self.dt
    [do stuff]
```
Some open questions:
- This requires that each module step to be self-contained, rather than the current `update_pre`, `update_post`, etc. I haven't figured out how to merge those two, or if we would even want to.
How would it work to have an intervention for example that is applied to two modules with different timesteps? When would it even get called?
- How do you aggregate results up to the sim in a way that makes sense?
- If we're going to implement all this infrastructure to handle uneven timesteps, should we explore an event-based implementation as well/instead???
- This also implies that we should move away from timesteps and instead use absolute time for everything, as mentioned in #482.

cliffckerr:
I suggest (responding also to #418) something like:
```py
# Sim in years
s1 = ss.Sim(start=1990, end=2040, dt=0.1, unit='year', diseases='hiv')

# Sim in days
s2 = ss.Sim(start='2025-01-01', end='2025-07-01', dt=7, unit='day', diseases='rsv')
```
[...]

PR #612, opened 2024-09-11
Rethink classes

### LINKED ISSUE #439: Rethink levels of inheritance
Modularity and reuse is good, right? Except it can get really confusing to figure out what is and isn't defined. Even for SIR, we have `Module → Disease → Infection → SIR`, with each one doing meaningful work. Pars and states and methods get seamlessly merged, which is great, but makes it quite opaque what's going on under the hood (an example was @jamiecohen's attempt to define a `rel_sus` property for HPV, which conflicted with a built-in `Infection` attribute).

Proposal: be more explicit about what is being called and initialized where. Specifically:
- [x] Each class should be self-contained in defining its own states, not relying on a deep stack of `super().__init__()` calls.
- [x] Method calls should be explicit, e.g. each class should define each critical method, even if it just calls a parent method (e.g. `def make_new_cases(self, sim): super().make_new_cases(sim)`)
- [x] Rethink `sim.step()` logic and if anything can be combined/simplified; likewise for disease methods
- [x] Think about if the sim can just be stored in each module, so it doesn't have to get passed around all the time

cliffckerr:
# Introduction

First major set of changes for Starsim v2.0: refactoring modules and the integration loop. See the changelog for details.

Closes #439, #601.

Some notes on the draft changes are below; some are out of date, see the changelog for the latest.

# Changes

## Integration loop

I changed integration loop in `sim.py` and individual modules to use `step()` almost everywhere instead of a mix of `update()` and custom names. I also combined/removed some steps. Specifically, the current `sim.py` integration loop is:
```py
demographics(): dem_mod.update()
diseases(): disease.update_pre()
connectors(): connector.update()
networks(): network.update()
interventions(): intervention(self)
diseases(): disease.make_new_cases()
people.resolve_deaths()
diseases(): disease.update_death(uids)
people.update_results()
demographics(): dem_mod.update_results()
diseases(): disease.update_results()
analyzers(): analyzer(self)
people.remove_dead()
people.update_post()
```
which I changed to:
```py
demographics(): dem.step()
diseases(): disease.step_pre()
networks(): network.step()
interventions(): intervention.step()
diseases(): disease.step()
people.check_deaths()
diseases(): disease.die(uids)
people.update_results()
mod(): modules.update_results()
analyzers(): analyzer.step()
people.finish_step()
```

Some notes:
- `disease.step_pre()` is optional; all updates can be done in `disease.step()` if this division isn't necessary. If different diseases (potentially) use different timesteps, `disease.step_pre()` is a bit dangerous anyway.
- `infection.make_new_cases()` is replaced by `infection.step()`, which calls `infection.infect()`, `infection.set_prognoses()`, and `infection.log_infections()`. `Disease` only has a `step()` method.

## Connectors

I moved some additional methods into `ss.Module`, which means it's now flexible enough to be the base class for pretty much everything. There is no more `ss.Connector`; instead, you can just insert a custom `ss.Module` anywhere you want it to appear in the integration. To reproduce the behavior of Starsim v1.0, where connectors are placed after demographics but before networks (somewhat arbitrarily), you could insert your connectors either at the end of the list of demographics or at the beginning of the list of networks. (In the past I've been in favor of an explicit `order` attribute for modules, but strong consensus was to use a pre-specified ordering instead.)

## States and transitions

It's not functional yet, but `sir.py` and `syphilis.py` show the syntax for how I'd like diseases to be designed. There is a new `ss.State` class, which is a special type of `ss.BoolArr`. There are also `ss.Event()` objects, which handle transitions between states. The idea is that these two things will handle all the boilerplate that's currently manual: ``ss.State`` will automatically create results tracking counts (unless asked not to); ``ss.Event()`` will handle all the timestep arrays (e.g. `ti_infected`), state transition checking, and results for transition counts (e.g. number of new infections). This should cut down on the amount of code required by a lot.

Currently, we have `default_pars()` and `add_states()`. These will be replaced by `define_pars()`, `define_states()`, `define_events()`, and `define_attrs()`. The last of these is for defining agent attributes that aren't states -- for example, `rel_sus` and `rel_trans`. I'm open to other names, but `define_` is the only one I could think of that works across all these methods and shows their similarities.

# Next steps

This is still WIP! These changes can still be discussed/rethought, so keen for feedback.

The next task I'll work on is time (#326), since that will likely impact how ``ss.Event`` is implemented. I hope to finish that by early next week.

Note that this PR is against the `rc2.0` branch; we won't merge any of this into `main` until all of the changes are complete (at least the big ones).


menriquez-IDM:
Cliff, I think using 'step' is a great implementation here is what I wanted also to suggest yesterday - 

image

this diagram is showing the conception of the step as an object where you should be able to make actions, and I put a couple of suggestions, but it may be more.

The section on top are properties and the ones below are methods.

starsim/diseases/sir.py
```diff
@@ -18,24 +18,67 @@ class SIR(ss.Infection):
...
             init_prev = ss.bernoulli(p=0.01),
             dur_inf = ss.lognorm_ex(mean=6),
             p_death = ss.bernoulli(p=0.01),
         )
         self.update_pars(pars, **kwargs)
-
-        self.add_states(
+        
+        self.define_states(
+            ss.BoolArr('susceptible', default=True, label='Susceptible'),
```
RomeshA, line 70: Aren't these already defined in `ss.Infection`? Why do they get recreated here?

RomeshA:
Broadly these changes look good to me. Two comments would be

- Just to confirm that the plan would be for events to coexist with the current style of implementing states (i.e., users would be able to choose either option, and/or mix-and-match when implementing their own diseases)? 
- I think we should have a different name for `Infection.infect()`, it would probably get confusing with the proposed `MyDisease.infect()` i.e., the event associated with infection that presumably would contain the code that is currently in `set_outcomes()` - `Infection.infect()`  is not about carrying out infections but is about determining who to infect. Presumably what would happen is that `set_outcomes()` could become `infect()` and then that function could be optionally registered as the callback for the infection event, for diseases that wanted to trigger it that way (and otherwise users could just call that function directly in `MyDisease.step()` the way it's currently implemented)

starsim/disease.py
```diff
@@ -58,13 +58,14 @@ def init_results(self):
...
+    def step_pre(self):
         """
-        Carry out autonomous updates at the start of the timestep (prior to transmission)
+        Carry out updates at the start of the timestep (prior to transmission);
+        these are typically state changes
         """
         pass
 
-    def update_death(self, uids):
+    def die(self, uids):
```
daniel-klein, line 68: 'die' is certainly better than 'update_death', but I wonder how it will be interpreted by users. What about something like 'handle_death' or 'on_death' to indicate that the agent is already dead and this is the chance to do any death-related cleanup?

starsim/modules.py
```diff
@@ -91,68 +102,57 @@ def update_pars(self, pars, **kwargs):
...
+    def finalize_results(self): # TODO: this is confusing, needs to be not redefined by the user, or called after a custom finalize_results()
         """ Finalize results """
         # Scale results
         for reskey, res in self.results.items():
             if isinstance(res, ss.Result) and res.scale:
                 self.results[reskey] = self.results[reskey]*self.sim.pars.pop_scale
         return
     
-    def add_states(self, *args, check=True):
+    def define_states(self, *args, check=True):
```
daniel-klein, line 155: Would there ever be a need for `update_states`, perhaps to allow a user to add states to a module that could then e.g. become the target of interventions? Might avoid the need to subclass in some cases?

starsim/diseases/sir.py
```diff
@@ -18,24 +18,67 @@ class SIR(ss.Infection):
...
+        )
+        
+        # Option 1 for defining events -- using objects
+        self.define_events(
+            ss.Event(self.susceptible, self.infected, func=self.infect,  reskey='infections'),
+            ss.Event(self.infected, self.recovered,   func=self.recover, reskey='recoveries'),
+            ss.Event(self.infected, self.dead,        func=self.die,     reskey='deaths'),
+        )
+        
+        # Option 2 for defining events -- using strings
```
daniel-klein, line 45: Option 2 feels less pythonic and could require additional checking.

starsim/diseases/sir.py
```diff
@@ -18,24 +18,67 @@ class SIR(ss.Infection):
...
+        self.define_events(
+            ss.Event('susceptible -> infected', self.infect,  reskey='infections'),
+            ss.Event('infected -> recovered',   self.recover, reskey='recoveries'),
+            ss.Event('infected -> dead',        self.die,     reskey='deaths'),
+        )
+        
+        self.define_attrs(
```
daniel-klein, line 52: I'm already struggling to see the functional distinction between states and attributes.

starsim/diseases/syphilis.py
```diff
@@ -54,7 +54,82 @@ def __init__(self, pars=None, **kwargs):
...
+            ss.State('latent_temp', label="Latent temporary"),  # Relapses to secondary (~1y)
+            ss.State('latent_long', label="Latent long"),  # Can progress to tertiary or remain here
+            ss.State('tertiary', label="Tertiary"),  # Includes complications (cardio/neuro/disfigurement)
+            ss.State('immune', label="Immune"),  # After effective treatment people may acquire temp immunity
+            ss.State('ever_exposed', label="Ever exposed", track_time=False),  # Anyone ever exposed - stays true after treatment
+            ss.State('congenital'),  # Congenital syphilis states
+        )
+        
+        self.define_events(
+            ss.Event(src='susceptible', dest=['exposed', 'infected', 'ever_exposed'], func=self.infect),
```
daniel-klein, line 72: Oh hmm, multiple destination states?

starsim/diseases/syphilis.py
```diff
@@ -54,7 +54,82 @@ def __init__(self, pars=None, **kwargs):
         )
         self.update_pars(pars, **kwargs)
 
-        self.add_states(
+        self.define_states(
+            # Adult syphilis states
+            ss.State('susceptible', label='Susceptible', default=True),
+            ss.State('exposed', label='Exposed'),  # AKA incubating. Free of symptoms, not transmissible
+            ss.State('primary', label='Primary'),  # Primary chancres
```
daniel-klein, line 61: Might be a good opportunity to use an enumeration as most of these are mutually exclusive, right?

starsim/disease.py
```diff
@@ -105,22 +106,27 @@ def set_prognoses(self, uids, source_uids=None):
         the log as part of this operation if logging is enabled (in the
         `Disease` parameters)
 
-        The `from_uids` are relevant for infectious diseases, but would be left
+        The `source_uids` are relevant for infectious diseases, but would be left
```
daniel-klein, line 109: Now called `sources`.

starsim/disease.py
```diff
@@ -105,22 +106,27 @@ def set_prognoses(self, uids, source_uids=None):
...
         `Disease` parameters)
 
-        The `from_uids` are relevant for infectious diseases, but would be left
+        The `source_uids` are relevant for infectious diseases, but would be left
         as `None` for NCDs.
 
         Args:
             sim (Sim): the STarsim simulation object
             uids (array): UIDs for agents to assign disease progoses to
             from_uids (array): Optionally specify the infecting agent
```
daniel-klein, line 115: Here also, `sources` rather than `from_uids`

starsim/disease.py
```diff
@@ -148,46 +154,22 @@ class Infection(Disease):
...
             ss.FloatArr('rel_sus', default=1.0, label='Relative susceptibility'),
             ss.FloatArr('rel_trans', default=1.0, label='Relative transmission'),
             ss.FloatArr('ti_infected', label='Time of infection' ),
         )
 
-        # Define random number generators for make_new_cases
-        self.rng_target = ss.rand_raw(name='target')
-        self.rng_source = ss.rand_raw(name='source')
+        # Define random number generators for step()
+        self.rng = ss.multi_random('target', 'source')
```
daniel-klein, line 166: Consider a more meaningful variable name.

starsim/disease.py
```diff
@@ -258,57 +260,37 @@ def make_new_cases(self):
...
+                p1p2b0 = [edges.p1, edges.p2, betamap[nk][0]]
+                p2p1b1 = [edges.p2, edges.p1, betamap[nk][1]]
+                for src, trg, beta in [p1p2b0, p2p1b1]:
+                    if beta: # Skip networks with no transmission
+    
+                        # Calculate probability of a->b transmission.
+                        beta_per_dt = net.beta_per_dt(disease_beta=beta, dt=self.sim.dt)
+                        p_transmit = rel_trans[src] * rel_sus[trg] * beta_per_dt
+        
+                        # Generate a new random number based on the two other random numbers -- 3x faster than `rvs = np.remainder(rvs_s + rvs_t, 1)`
```
daniel-klein, line 281: Remove performance comment about an algorithm we no longer use

starsim/disease.py
```diff
@@ -258,57 +260,37 @@ def make_new_cases(self):
...
-            new_cases = np.empty(0, dtype=int)
-            sources = np.empty(0, dtype=int)
-            networks = np.empty(0, dtype=int)
-
-        if len(new_cases):
-            self._set_cases(new_cases, sources)
-            
+        new_cases = ss.uids.cat(new_cases)
+        sources = ss.uids.cat(sources)
+        networks = None # TEMP, NEED TO FIX
```
daniel-klein, line 290: ALL CAPS, must be important!

starsim/distributions.py
```diff
@@ -972,7 +973,58 @@ def __init__(self, values=None, bins=None, density=False, data=None, **kwargs):
...
+        n_args = len(args)
+        n_dists = len(self)
+        if n_args != len(self):
+            errormsg = f'Number of UID lists supplied ({n_args}) does not match number of distributions ({n_dists})'
+            raise ValueError(errormsg)
+        
+        # Generate the random numbers
+        rvs = self.dists[0].rvs(args[0]) # Get the initial random numbers
+        for i,uids in enumerate(args[1:]):
+            rvs += self.dists[i+1].rvs(uids)
```
daniel-klein, line 1016: WAIT, what?? Why are we adding and taking the modulus?! This algorithm introduces significant bias in application and should be removed in favor of the newer "bit mixing" approaches that have similar performance characteristics.

starsim/interventions.py
```diff
@@ -399,7 +373,7 @@ def init_pre(self, sim):
         self.outcomes = {k: np.array([], dtype=int) for k in ['unsuccessful', 'successful']} # Store outcomes on each timestep
         return
 
-    def get_accept_inds(self, sim):
+    def get_accept_inds(self, sim): # TODO: do not pass sim
```
daniel-klein, line 376: Reminder to address TODO

starsim/modules.py
```diff
@@ -228,14 +239,3 @@ def plot(self):
                 ax.set_title(k)
                 ax.set_xlabel('Year')
         return fig
-
-
-class Connector(Module):
```
daniel-klein, line 233: I still prefer to have the concept of a Connector formally as part of the framework, even if it is just a veneer. It gives legitimacy to the feature and a clear building block for users.

starsim/modules.py
```diff
@@ -39,13 +39,13 @@ def find_modules(key=None):
             except:
                 pass
     return modules if key is None else modules[key]
-
+    
 
 class Module(sc.quickobj):
 
-    def __init__(self, name=None, label=None, requires=None):
-        self.set_metadata(name, label, requires) # Usually reset as part of self.update_pars()
```
daniel-klein, line 47: Wait, I though requires was a nice thing to have to ensure compatibility. Why remove?

starsim/disease.py
menriquez-IDM: I agree with Dan,  that being said, some observations:
_handle_death_ : feels like it covers everything you'd want to take care of when an agent dies (managing all tasks related). 
_on_death_ :  this kind of fits the event-driven programming, signaling that this is triggered when the agent dies, which would be more intuitive for users/developers that are more familiar with this event-driven programming style. 

I would even suggest a combination of both : _on_death_cleanup()_

starsim/sim.py
```diff
@@ -180,87 +188,77 @@ def init_results(self):
...
         # Carry out autonomous state changes in the disease modules. This allows autonomous state changes/initializations
         # to be applied to newly created agents
         for disease in self.diseases():
-            disease.update_pre()
+            if isinstance(disease, ss.Disease): # Could be a connector instead -- TODO, rethink this
+                disease.step_pre()
 
-        # Update connectors -- TBC where this appears in the ordering
-        for connector in self.connectors():
-            connector.update()
```
daniel-klein, line 220: So as I understand, the current functionality of connectors is lost with this refactor as they updated after disease pre and before network transmission.

starsim/states.py
```diff
@@ -352,6 +362,11 @@ def split(self):
         f_uids = self.false()
         return t_uids, f_uids
 
+
+class State(BoolArr):
```
daniel-klein, line 366: I understood State to be a _future_ feature not included in this PR? Relabeling a BoolArr does not seem worthwhile at this point.

starsim/utils.py
```diff
@@ -189,9 +177,19 @@ def find_contacts(p1, p2, inds):  # pragma: no cover
     return pairing_partners
 
 
-# %% Seed methods
+def check_requires(sim, requires, *args):
```
daniel-klein, line 180: Confused as to the purpose here if one module (e.g. Connector) can no longer explicitly state that it requires other modules.

tests/test_plugins.py
```diff
@@ -93,15 +95,15 @@ def test_connectors(do_plot=False):
     sims = sc.objdict() # List of sims
     
     # Make a sim with a connector, and run
-    sims.con = ss.Sim(label='With connector', connectors=hiv_syph(), **args)
+    sims.con = ss.Sim(label='With connector', interventions=hiv_syph(), **args)
```
daniel-klein, line 98: Nope, I don't like framing these as "interventions"

daniel-klein (commented):
Good overall. Mostly concerned about the new multi_rand using the debunked modulus approach and lack of connectors.

docs/run_tutorials
```diff
@@ -0,0 +1,62 @@
...
+"""
+
+import os
+import sciris as sc
+import nbformat
+import nbconvert.preprocessors as nbp
+
+folder = './tutorials'
+timeout = 600
+yay = '✓'
```
pausz, line 19: :partying_face: (yay!)

starsim/diseases/sir.py
cliffckerr: The idea is to have an optional `inherit` parameter, which does or doesn't inherit states from previous classes. In any case, I think states can be redefined without error. Need to check/fix this though since it's not clear how it currently works or how it should work!

starsim/disease.py
cliffckerr: I think in general we need to do a better job of differentiating between methods that schedule things to happen, and methods that actually make them happen. Have thought about it some, need to think about it more!

starsim/modules.py
cliffckerr: You can define new states with `define_states`, so there isn't a need for a separate `update_states`. But perhaps we should just go with `add` for all of them?

starsim/diseases/sir.py
cliffckerr: Agree, users would be able to choose (maybe -- we'll see if option 2 makes sense!).

starsim/diseases/sir.py
cliffckerr: States get automatically counted in results, etc., while attributes don't. States are also only ever `BoolArr` (and maybe `IntArr` if we implement that).

starsim/diseases/syphilis.py
cliffckerr: Exactly! This was a big downside of Dymodetron.

starsim/disease.py
```diff
@@ -258,57 +260,37 @@ def make_new_cases(self):
...
+                for src, trg, beta in [p1p2b0, p2p1b1]:
+                    if beta: # Skip networks with no transmission
+    
+                        # Calculate probability of a->b transmission.
+                        beta_per_dt = net.beta_per_dt(disease_beta=beta, dt=self.sim.dt)
+                        p_transmit = rel_trans[src] * rel_sus[trg] * beta_per_dt
+        
+                        # Generate a new random number based on the two other random numbers -- 3x faster than `rvs = np.remainder(rvs_s + rvs_t, 1)`
+                        randvals = self.rng.rvs(src, trg)
+                        transmitted = p_transmit > randvals
```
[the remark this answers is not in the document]
cliffckerr: I always used to write it like that, but to me it makes more sense like this. Can try to explain why if you want, I don't have a clear explanation at the moment :)

starsim/disease.py
cliffckerr: Yes I just deleted this without fixing it -- definitely need to fix prior to merge, even for `rc2.0`!

starsim/distributions.py
cliffckerr: Ah good catch! I think this was an accidental regression.

starsim/modules.py
cliffckerr: Good point for discussion!

starsim/modules.py
cliffckerr: The main reason is that it was hard-coded to specific disease names. So if your module required `'hiv'`, but you subclassed that to `hiv2`, the requirement would fail. There is still a `check_requires` function, which is what `requires` called under the hood, so the idea is that this could still be used directly in modules if desired.

starsim/sim.py
cliffckerr: Good points -- let's discuss!

starsim/states.py
cliffckerr: Agree, this is to enable the automated transitions refactor -- it doesn't too much yet.

starsim/disease.py
cliffckerr: Renamed `step_die()`

starsim/disease.py
cliffckerr: Fixed

starsim/interventions.py
cliffckerr: Fixed -- this was annoyingly hard! Show how important it is to get the API right the first time.

starsim/modules.py
cliffckerr: Restored

starsim/utils.py
cliffckerr: It can, it's just slightly more manual now.

tests/test_plugins.py
cliffckerr: Restored

starsim/disease.py
```diff
@@ -148,46 +154,22 @@ class Infection(Disease):
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
-        self.add_states(
+        self.define_states(
             ss.BoolArr('susceptible', default=True, label='Susceptible'),
```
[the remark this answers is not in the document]
cliffckerr: You can call `define_states()` more than once if you want to redefine a state, but I do kind of like keeping this name as I think this usage should be discouraged.

starsim/diseases/sir.py
```diff
@@ -18,24 +18,67 @@ class SIR(ss.Infection):
...
+        
+        self.define_states(
+            ss.State('susceptible', default=True),
+            ss.State('infected'),
+            ss.State('recovered'),
+        )
+        
+        # Option 1 for defining events -- using objects
+        self.define_events(
+            ss.Event(self.susceptible, self.infected, func=self.infect,  reskey='infections'),
```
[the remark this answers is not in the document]
cliffckerr: There will probably be a new module for it

starsim/disease.py
```diff
@@ -43,7 +42,8 @@ def _boolean_states(self):
     def init_pre(self, sim):
         """ Link the disease to the sim, create objects, and initialize results; see Module.init_pre() for details """
         super().init_pre(sim)
-        self.init_results()
+        if self.pars.log:
+            self.log = InfectionLog()
         return
 
     def init_results(self):
```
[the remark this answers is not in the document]
cliffckerr: Yes, I think we've gone back and forth on this. In Covasim, this was an analyzer.

PR #623, opened 2024-09-13
Rc2.0 rethink time

cliffckerr:
PR done, still working on changelog, tests, and documentation though!


starsim/parameters.py
```diff
@@ -282,17 +319,17 @@ def validate_dt(self):
...
-                # Round the reciprocal
-                reciprocal = int(reciprocal)
-                rounded_dt = 1.0 / reciprocal
-                self.dt = rounded_dt
-                if self.verbose:
-                    warnmsg = f'Warning: Provided time step dt={dt} resulted in a non-integer number of steps per year. Rounded to {rounded_dt}.'
-                    ss.warn(warnmsg)
+        dt = self.dt
+        reciprocal = 1.0 / dt  # Compute the reciprocal of dt
+        if not reciprocal.is_integer():  # Check if reciprocal is not a whole (integer) number
```
daniel-klein, line 324: Does this matter? On each time step, we will increment `ti` and add `dt` to the current time, all while accounting for units in durations and rates. This question becomes more complicated if we get to the point of having different parts of the model update at different frequencies.

starsim/diseases/sir.py
```diff
@@ -145,10 +145,11 @@ class SIS(ss.Infection):
...
         super().__init__()
         self.define_pars(
-            beta = 0.05,
+            unit = 'year',
+            beta = ss.rate(0.05),
             init_prev = ss.bernoulli(p=0.01),
-            dur_inf = ss.lognorm_ex(mean=10),
-            waning = 0.05,
+            dur_inf = ss.dur(ss.lognorm_ex(mean=10)),
+            waning = ss.rate(0.05),
```
daniel-klein, line 152: Fixed rates implicitly create memoryless compartmental-like dynamics. But in agent-based modeling, there's no need for geometrically-distributed waiting times. These waiting times are the discrete-time analog of exponential waiting times in an ODE. Are we doing a disservice by facilitating the use of `rate`? One easy one to implement could be a weibull, the rate of which depends lineary on waiting time. In fact, is `ss.rate(0.05)` any different from `ss.dur(ss.exponential(0.05))`? Might it be possible to collapse everything to distributions and assume `ss.dur`?

From a continuous distribution of duration, we could sample ONE time, set a timer, and wait. ALTERNATIVELY, we could compute the transition probability on EACH time step from the cumulative, after conditioning on not already transitioning, I suppose. If we do the math right, the exponential should always return the same transition prob.

The key thing to think through here is performance. If the duration is unlikely to change _while waiting_, draw ONCE and set a timer. However, if other dynamics/interventions are continually modifying the probability (as in TB), we might be better off with the EACH step approach. And actually, there's some possible trickery here to think through with respect to changing distribution parameters on the fly. So perhaps the framework could faciliate here?

starsim/sim.py
```diff
@@ -131,7 +131,8 @@ def initialize(self, **kwargs):
     def init_time_attrs(self):
         """ Time indexing; derived values live in the sim rather than in the pars """
         self.dt = self.pars.dt # Shortcut to dt since used a lot
-        self.yearvec = np.arange(start=self.pars.start, stop=self.pars.end + self.pars.dt, step=self.pars.dt) # The time points of the sim
+        stop = self.pars.date_add(self.pars.end, self.pars.dt) # Potentially convert to a date
+        self.yearvec = np.arange(start=self.pars.start, stop=stop, step=self.pars.dt) # The time points of the sim
```
daniel-klein, line 135: Consider using `datetime` from here? Also for `sim.t`?

starsim/sim.py
```diff
@@ -287,7 +288,9 @@ def run(self, until=None, verbose=None):
             elapsed = T.toc(output=True)
             if verbose: # Print progress
                 simlabel = f'"{self.label}": ' if self.label else ''
-                string = f'  Running {simlabel}{self.yearvec[self.ti]:0.1f} ({self.ti:2.0f}/{self.npts}) ({elapsed:0.2f} s) '
+                timepoint = self.yearvec[self.ti]
+                timelabel = f'{timepoint:0.1f}' if isinstance(timepoint, float) else str(timepoint) # TODO: fix
```
daniel-klein, line 292: Datetime has nice stringification

tests/test_time.py
```diff
@@ -0,0 +1,47 @@
...
+    
+    pars = dict(
+        diseases = 'sis',
+        networks = 'random',
+        n_agents = small,
+    )
+    
+    sims = sc.objdict()
+    sims.y = ss.Sim(pars, unit='year', label='Year', start=2000, end=2002, dt=1/365)
+    sims.d = ss.Sim(pars, unit='day', label='Day', start='2000-01-01', end='2002-01-01', dt=1)
```
daniel-klein, line 26: dt=1 should be the default, yeah?

daniel-klein: Wait, now I'm confused - does simulation-level units cascade down to other modules>?

daniel-klein:
I've been thinking through ramifications of modules having differing time steps. Some things get tricky. For example, when everything is synchronous, I can reset multipliers at the beginning (end) of each time step and recalculate every update. But such interactions may not work if adjacent modules are updating at different frequencies. I don't see this draft PR enabling differing time steps yet, just different time UNITS, but something to think about as we move ahead.

starsim/parameters.py
cliffckerr: TBC, but I think it could still be an issue for the reciprocal to be not-an-integer. Users will have the opportunity to choose different time units, so the reciprocal just has to be an integer for their chosen time unit (for example, you could put 0.5 but not 0.7 days as your time unit).

starsim/diseases/sir.py
cliffckerr: "Duration" does not necessarily assume an exponential distribution; it specifies how the quantity scales with time. For example, you could specify "time_to_symptoms" as a normal distribution with a mean of 20 days and a standard deviation of 3 days; this would be `ss.dur(ss.normal(20,3))`. What `ss.dur()` is doing is specifying that if the time unit is actually in timesteps of `dt=2`, that this would be correctly translated to `ss.normal(10,1.5)` in timestep-units. Happy to consider other APIs, though!

starsim/parameters.py
```diff
@@ -168,12 +168,12 @@ def __init__(self, **kwargs):
         self.pop_scale = None  # How much to scale the population
 
         # Simulation parameters
-        self.unit       = 'year' # The time unit to use (NOT YET IMPLEMENTED)
+        self.unit       = 'year' # The time unit to use; options are None, 'year' (default), and 'day'
```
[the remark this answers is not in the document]
cliffckerr: I think for "abstract" modules like SIR, the default units would be `None`, and it would take on the Sim unit without changing the parameter values. For non-abstract modules like `HIV`, it would have units, and if the user specified a different unit for the sim (e.g. `day`), the default would be to translate the parameters to that timestep, but the user would also have the option to leave (or independently modify) the module-level timestep.

starsim/sim.py
cliffckerr: The problem with `datetime` is that it's so ... date-based! I can imagine various Starsim use cases where it doesn't make sense to have a fixed start date: for example, you might have "Weeks since start of outbreak" as your x-axis, and it would be odd to be forced to specify a precise date.

starsim/sim.py
cliffckerr: Yes ... maybe this is ok actually!

tests/test_time.py
cliffckerr: Yes, this is just to be explicit

tests/test_time.py
```diff
@@ -0,0 +1,47 @@
...
+    
+    for sim in sims.values():
+        sim.run()
+        
+    rtol = 0.01
+    # assert np.isclose(sims.y.summary.cum_infections, sims.d.summary.cum_infections, rtol=rtol)
```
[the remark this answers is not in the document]
cliffckerr: I will once it works :)

cliffckerr:
Exactly -- this is a first step towards allowing different modules to have different timesteps. I think it's an open question about how much to hold the user's hand if they specify non-aligned timesteps for different modules.

starsim/disease.py
```diff
@@ -276,7 +298,8 @@ def infect(self):
                     if beta: # Skip networks with no transmission
     
                         # Calculate probability of a->b transmission.
-                        beta_per_dt = net.beta_per_dt(disease_beta=beta, dt=self.sim.dt)
+                        beta_per_dt = edges.beta * beta # TODO: think of shortcut to skip this multiplication if all edges are just 1.0
```
daniel-klein, line 301: Previously, we asked the network to compute transmissibility. For example, STI networks would use act rates and condom use. Do we now lose that functionality?

starsim/distributions.py
```diff
@@ -21,25 +21,13 @@ def str2int(string, modulo=1_000_000):
...
-
 def link_dists(obj, sim, module=None, overwrite=False, init=False, **kwargs):
     """ Link distributions to the sim and the module; used in module.initialize() and people.initialize() """
     if module is None and isinstance(obj, ss.Module):
         module = obj
-    dists = ss.find_dists(obj, **kwargs) # Important that this comes first, before the sim is linked to the dist!
+    dists = ss.find_objs(Dist, obj, **kwargs) # Important that this comes first, before the sim is linked to the dist!
     for key,val in dists.items():
-        if isinstance(val, ss.Dist):
+        if isinstance(val, ss.Dist): # TODO: should be able to skip this if statement since should always be true?
```
daniel-klein, line 30: Yeah, agree with the comment

starsim/diseases/sir.py
```diff
@@ -145,10 +144,10 @@ class SIS(ss.Infection):
     def __init__(self, pars=None, *args, **kwargs):
         super().__init__()
         self.define_pars(
-            beta = 0.05,
+            beta = ss.rate(0.05),
             init_prev = ss.bernoulli(p=0.01),
-            dur_inf = ss.lognorm_ex(mean=10),
-            waning = 0.05,
+            dur_inf = ss.lognorm_ex(mean=ss.dur(10)), # TODO: handle gotcha of mean=ss.dur(10), std=3 -- should be able to automate it?
```
daniel-klein, line 149: Is it possible for these parameters to have units? For example, `ss.dur(10, 'weeks')`? Same with rates, `ss.rate(2, 'per year')` or similar?

starsim/modules.py
```diff
@@ -69,10 +70,16 @@ def disp(self, output=False):
...
         """ Set metadata for the module """
-        self.name = sc.ifelse(name, getattr(self, 'name', self.__class__.__name__.lower())) # Default name is the class name
-        self.label = sc.ifelse(label, getattr(self, 'label', self.name))
+        self.name  = sc.ifelse(name,  getattr(self, 'name',  self.pars.get('name', self.__class__.__name__.lower()))) # Default name is the class name
+        self.label = sc.ifelse(label, getattr(self, 'label', self.pars.get('label', self.name)))
+        return
+    
+    def set_time_pars(self, unit=None, dt=None):
+        """ Set time units for the module """
+        self.unit  = sc.ifelse(unit,  getattr(self, 'unit', self.pars.get('unit')))
```
daniel-klein, line 81: Does this mean that each module has a single `unit` that is applied to everything in that module? Can we not mix units? `mean1=ss.dur(3, 'days')`, `mean2=ss.dur(5, 'weeks')`?

starsim/network.py
```diff
@@ -300,9 +300,9 @@ def remove_uids(self, uids):
 
         return
 
-    def beta_per_dt(self, disease_beta=None, dt=None, uids=None):
-        if uids is None: uids = Ellipsis
-        return self.edges.beta[uids] * disease_beta * dt
+    # def beta_per_dt(self, disease_beta=None, dt=None, uids=None):
+    #     if uids is None: uids = Ellipsis
+    #     return self.edges.beta[uids] * disease_beta * dt
```
daniel-klein, line 305: I thought this was a good idea, and was considering using this pattern for a new network based on mixing pools.

starsim/parameters.py
```diff
@@ -258,43 +258,81 @@ def validate_total_pop(self):
...
     
-    def validate_start_end(self):
-        """ Ensure at least one of n_years and end is defined, but not both """
+    def int_to_date(self, x):
+        """ Convert an integer to a date """
+        if self.unit == 'year':
+            date = sc.date(f'{x}-01-01')
+        else:
+            raise NotImplementedError
+        return date
```
daniel-klein, line 267: Would it make sense for these utility functions to live in `utils.py`?

starsim/parameters.py
```diff
@@ -258,43 +258,81 @@ def validate_total_pop(self):
...
+        if sc.isnumber(start) and sc.isnumber(end):
+            dur = end - start
+        else:
+            if self.unit == 'year':
+                dur = sc.datetoyear(end) - sc.datetoyear(start) # TODO: allow non-integer amounts
+            elif self.unit == 'day':
+                dur = (end - start).days
+            else:
+                raise NotImplementedError
+        return dur
```
daniel-klein, line 293: Can we just do everything in datetime and skip this mess?

starsim/parameters.py
```diff
@@ -258,43 +258,81 @@ def validate_total_pop(self):
...
+            if self.dur is not None:
+                self.end = self.date_add(self.start, self.dur)
             else:
-                errormsg = 'You must supply one of n_years and end."'
+                errormsg = 'You must supply either "dur" or "end".'
                 raise ValueError(errormsg)
         return
 
-    def validate_dt(self):
+    def validate_dt(self): # TODO: may not be needed
```
daniel-klein, line 318: Consider removing, I do not want to model to change my dt and see no reason why it should.

starsim/disease.py
cliffckerr: It should still be doable, I might need to look at STIsim to see how to translate it though!

starsim/people.py
```diff
@@ -277,7 +277,7 @@ def update_post(self):
         """ Final updates at the very end of the timestep """
         sim = self.sim
         if sim.pars.use_aging:
-            self.age[self.alive.uids] += sim.dt
+            self.age[self.alive.uids] += sim.dt_year
```
daniel-klein, line 280: This is probably right. Confusing at first, but makes sense when you think about it. To avoid the "think about it", maybe `sim.dt(units='year')`? OR, could this be `ss.rate(1)`? Is that better? People get older at a rate of 1 year per year.

starsim/diseases/sir.py
cliffckerr: Yes, unit is indeed the second argument!

starsim/modules.py
cliffckerr: The module's `unit` is used for its timestep, but the parameters themselves can have different units.

starsim/sim.py
daniel-klein: I still think maintaining time internally within the model (e.g. `timevec`) as a `datetime` could provide benefits, including stringification and easy deltas.

starsim/network.py
cliffckerr: All the same code is there, just moved around. But agree we might need to think more carefully about this.

starsim/sim.py
```diff
@@ -630,7 +652,7 @@ def plot(self, key=None, fig=None, style='fancy', fig_kw=None, plot_kw=None):
                 
             # Do the plotting
             for ax, (key, res) in zip(axs, flat.items()):
-                ax.plot(yearvec, res, **plot_kw, label=self.label)
+                ax.plot(timevec, res, **plot_kw, label=self.label)
```
daniel-klein, line 655: Another benefit of datetime is that matplotlib and seaborn know how to handle it nicely in plots!

starsim/parameters.py
cliffckerr: Yes, I've moved them to `time.py`

starsim/parameters.py
cliffckerr: It's still a mess because dates are just inherently messy! You can't do math with dates, for example -- if you have `unit='year'` and `dt=0.5`, good luck getting `datetime` to understand that. If we were committing to `unit='day'` and `dt` forced to be an integer, then it would work -- but everything else would still be complicated.

starsim/parameters.py
cliffckerr: Removed

starsim/people.py
cliffckerr: This is just a quirk of how we think of ages. Even if the simulation timestep is days, we still want people to age in years. Same with some other things (e.g. birth rate, which is a rate per year regardless of timestep and unit).

starsim/sim.py
cliffckerr: Yes -- if `start` and `end` are dates, then it will use `datetime`. But we have to agree to disagree about the default plotting -- I am not a fan!
```py
import numpy as np
import sciris as sc
import pylab as pl

x = sc.daterange('2020-01-01', '2023-01-01', as_date=True)
y = sc.rolling(np.random.rand(len(x)), window=30)

pl.subplot(2,1,1)
pl.plot(x,y)

pl.subplot(2,1,2)
pl.plot(x,y)
sc.dateformatter()
```
image

starsim/utils.py
```diff
@@ -159,6 +159,18 @@ def warn(msg, category=None, verbose=None, die=None):
     return
 
 
+def find_objs(cls, obj, verbose=False, **kwargs):
```
daniel-klein, line 162: Consider saying what `skip` does here, as I see that argument used in a call from `distributions.py`

starsim/time.py
```diff
@@ -0,0 +1,157 @@
...
+    
+#%% Helper functions
+
+# Define available time units
+time_units = sc.dictobj(
+    day = 1,
+    week = 7,
+    month = 30.4375, # 365.25/12 -- more accurate and nicer fraction
+    year = 365, # For simplicity with days
+)
```
daniel-klein, line 20: If using datetime, there's alreay definitions for year, month, day, hour, minute, second, and microsecond.

starsim/sim.py
```diff
@@ -130,14 +130,28 @@ def initialize(self, **kwargs):
     
     def init_time_attrs(self):
         """ Time indexing; derived values live in the sim rather than in the pars """
-        self.dt = self.pars.dt # Shortcut to dt since used a lot
-        self.yearvec = np.arange(start=self.pars.start, stop=self.pars.end + self.pars.dt, step=self.pars.dt) # The time points of the sim
-        self.results.yearvec = self.yearvec # Store the yearvec in the results for plotting
-        self.npts = len(self.yearvec) # The number of points in the sim
+        stop = self.pars.date_add(self.pars.end, self.pars.dt) # Potentially convert to a date
+        if sc.isnumber(self.pars.start):
+            self.timevec = np.arange(start=self.pars.start, stop=stop, step=self.pars.dt) # The time points of the sim
```
daniel-klein, line 135: dt.calendar?

starsim/sim.py
```diff
@@ -130,14 +130,28 @@ def initialize(self, **kwargs):
...
+                if self.pars.dt < 1:
+                    errormsg = f'Cannot use a timestep of less than a day ({self.pars.dt}) with date-based indexing'
+                    raise ValueError(errormsg)
+                day_delta = int(np.round(self.pars.dt))
+            self.timevec = sc.daterange(self.pars.start, self.pars.end, interval={'days':day_delta})
+        self.results.timevec = self.timevec # Store the yearvec in the results for plotting
+        self.npts = len(self.timevec) # The number of points in the sim
         self.tivec = np.arange(self.npts) # The vector of time indices
         self.ti = 0  # The time index, e.g. 0, 1, 2
+        self.dt_year = ss.time_ratio(self.pars.unit, self.pars.dt, 'year', 1.0) # Figure out what dt is in years; used for demographics # TODO: handle None
```
daniel-klein, line 152: All this time handling is messy to do on our own. It would help to use an existing library, datetime is the one I know.

starsim/time.py
```diff
@@ -0,0 +1,157 @@
...
+
+def time_ratio(unit1='day', dt1=1.0, unit2='day', dt2=1.0):
```
daniel-klein, line 22: Hoping we won't need this function

starsim/time.py
```diff
@@ -0,0 +1,157 @@
...
+
+
+#%% Time classes
+
+class TimeUnit:
+    """ Base class for durations and rates """
+    def __init__(self, value, unit=None, self_dt=None, parent_unit=None, parent_dt=None):
+        self.value = value
+        self.unit = unit
+        self.self_dt = self_dt
```
daniel-klein, line 62: Why not just `dt`?

starsim/time.py
```diff
@@ -0,0 +1,157 @@
...
+    def __repr__(self):
+        name = self.__class__.__name__
+        return f'ss.{name}({self.value}, unit={self.unit}, dt={self.self_dt})'
+    
+    def disp(self):
+        return sc.pr(self)
+    
+    def set_factor(self):
+        """ Set factor used to multiply the value to get the output """
+        self.factor = time_ratio(unit1=self.unit, dt1=self.self_dt, unit2=self.parent_unit, dt2=self.parent_dt)
```
daniel-klein, line 105: So factor is in relation to my immediate partent, which might have its own factor in relation to other (higher) units?

daniel-klein: Ah but in practice, is unit2 always at the simulation level? Confused...

starsim/time.py
```diff
@@ -0,0 +1,157 @@
...
+        return self.value/self.factor
+    
+
+class time_prob(TimeUnit):
+    """ A probability over time (a.k.a. a "true" rate, cumulative hazard rate); must be >0 and <1 """
+    @property
+    def x(self):
+        numer = (1 - np.exp(-self.value/self.factor))
+        denom = (1 - np.exp(-self.value))
+        return self.value*numer/denom
```
daniel-klein, line 155: Wait, why not just return `numer`?

daniel-klein (commented):
Consider building time-aware Starim off existing libraries. For example, internally, time could be handled using `datetime`, which provides differencing, stringification, etc.

Another thought would be to build off a package (or borrow ideas) from the `units` package (https://pypi.org/project/units/). 

```python
from units import unit
metre = unit('m')
print(metre(7) + metre(11))
```
Could we have parameters like this? E.g. `mean=months(3.2)`? Combined with a datetime calendar, this could work?

starsim/time.py
cliffckerr: Because when we specify a rate, it already has time units associated with it. For example, say `r=0.1` in units of `dt=1`. If the simulation timestep is also `dt=1`, then `self.factor=1`, but we have `1 - np.exp(-0.1) = 0.095`.

But you're right that this solution is also wrong :D Here's the corrected version:
```py
        rate = -np.log(1 - self.value)
        out = 1 - np.exp(-rate/self.factor)
```

starsim/time.py
cliffckerr: There are, but not ones that can be used mathematically, as far as I know (e.g., as would be used to convert `ss.dur(6, unit='month')` into timesteps of `dt=0.1, unit='year'`).

starsim/sim.py
cliffckerr: I think the `calendar` module is for something different, no?

starsim/sim.py
cliffckerr: It is messy, but there really isn't a commonly used package to do this. A lot of utilities for working with `datetime` are in Sciris, so those are being leveraged where possible. But here, for example, I don't think there's a `datetime` equivalent.

starsim/time.py
cliffckerr: Because in practice, it would be very rare to specify `self_dt`, and common to specify `parent_dt`. Even though I wrote this class, I myself got confused by swapping the two accidentally! Typically when you specify a parameter, it would be something like `ss.dur(4, unit='year')`. It's an anti-pattern to specify the equivalent `ss.dur(2, unit='year', self_dt=2)`. However, you often _would_ specify the parent timestep, i.e. `ss.dur(4, unit='year', parent_unit='year', parent_dt=0.1)`, although outside of manual testing this initialization would typically be done by the sim/module automatically.

cliffckerr: Would be open to just removing `self_dt`, however!

starsim/time.py
cliffckerr: Correct, the `parent` is the module doing the integration, whose timestep may or may not match the simulation. For example, a parameter `birth_rate = ss.rate(0.05, 'year')` might reside inside `mod = ss.Births(unit='year', dt=1.0)`, which itself exists within `ss.Sim(unit='day', dt=7)`. Since `birth_rate` and `mod` have the same unit, the rate of 0.05 is applied at the timestep, with no conversion to the sim time units necessary.

cliffckerr:
I like the idea of using an existing library, but there don't seem to be any commonly used/good ones except for `datetime` and `dateutils` (and `sciris` :grin: , which builds on both of those). To clarify, the current implementation _is_ built around `datetime` -- specifically, if you specify `start` as a datetime (or date-like string), then `timevec` will also be a datetime array.

`NumPy` and `pandas` also both have native datetime handling, but both are based around nanosecond timestamps, which are clunky for our purposes.

I wouldn't mind something like `mean=ss.months(3.2)` as a shortcut for `mean=ss.dur(3.2, 'month')`, but this would be confusing for rates (I guess we could have `ss.permonth()` etc?).

tests/test_time.py
```diff
@@ -0,0 +1,139 @@
...
+    # Test duration dt
+    d1 = ss.dur(2)
+    d2 = ss.dur(3)
+    d3 = ss.dur(2, parent_dt=0.1)
+    d4 = ss.dur(3, parent_dt=0.2)
+    for d in [d1,d2,d3,d4]: d.init()
+    
+    assert d1 + d2 == 2+3
+    assert d3 + d4 == 2/0.1 + 3/0.2
+    assert d3 * 2 == 2/0.1*2
```
RomeshA, line 41: These could be used to do useful conversions e.g., multiplying a rate by a dur to get a number (like 'tests per year' multiplied by a timestep that's a number of day)

starsim/diseases/ebola.py
```diff
@@ -19,17 +19,17 @@ def __init__(self, pars=None, *args, **kwargs):
...
             unburied_factor = 2.1,
             
             # Natural history parameters, all specified in days
-            dur_exp2symp    = ss.lognorm_ex(mean=12.7), # Add source
-            dur_symp2sev    = ss.lognorm_ex(mean=6), # Add source
-            dur_sev2dead    = ss.lognorm_ex(mean=1.5), # Add source
-            dur_dead2buried = ss.lognorm_ex(mean=2), # Add source
-            dur_symp2rec    = ss.lognorm_ex(mean=10), # Add source
-            dur_sev2rec     = ss.lognorm_ex(mean=10.4), # Add source
+            dur_exp2symp    = ss.lognorm_ex(mean=ss.days(12.7)), # Add source
```
RomeshA, line 27: This looks great

starsim/diseases/syphilis.py
```diff
@@ -349,12 +345,13 @@ def set_congenital(self, uids, source_uids=None):
...
 
                 # Schedule events
+                ratio = ss.time_ratio(unit1='year', dt1=1.0, unit2=self.unit, dt2=self.dt) # TODO: think about simplifying
                 for oi, outcome in enumerate(self.pars.birth_outcome_keys):
                     o_uids = state_uids[assigned_outcomes == oi]
                     if len(o_uids) > 0:
                         ti_outcome = f'ti_{outcome}'
                         vals = getattr(self, ti_outcome)
-                        vals[o_uids] = sim.ti + rr(time_to_birth[o_uids] / sim.dt)
+                        vals[o_uids] = sim.ti + rr(time_to_birth[o_uids] * ratio)
```
RomeshA, line 354: Ideally instead of `time_to_birth[o_uids] * ratio` this would be `int(time_to_birth[o_uids]/self.dt)`?

starsim/sim.py
```diff
@@ -115,29 +139,29 @@ def initialize(self, **kwargs):
...
     
     def init_time_attrs(self):
         """ Time indexing; derived values live in the sim rather than in the pars """
-        self.dt = self.pars.dt # Shortcut to dt since used a lot
-        self.yearvec = np.arange(start=self.pars.start, stop=self.pars.end + self.pars.dt, step=self.pars.dt) # The time points of the sim
-        self.results.yearvec = self.yearvec # Store the yearvec in the results for plotting
-        self.npts = len(self.yearvec) # The number of points in the sim
-        self.tivec = np.arange(self.npts) # The vector of time indices
+        pars = self.pars
+        self.timevec = ss.make_timevec(pars.start, pars.stop, pars.dt, pars.unit)
```
RomeshA, line 157: Not `tvec`? In STIsim we have a `tivec` so this would complement that (and I think we could also have a `sim.t` for the current time which returns `sim.tvec[sim.ti]`)

starsim/sim.py
```diff
@@ -115,29 +139,29 @@ def initialize(self, **kwargs):
...
+        self.results.timevec = self.timevec # Store the timevec in the results for plotting # TODO: instead, store a timevec with each result
+        self.npts = len(self.timevec) # The number of points in the sim
+        self.timearray = np.arange(self.npts)*pars.dt # Absolute time array
```
RomeshA, line 160: Is `self.timearray` for performance?

starsim/sim.py
```diff
@@ -115,29 +139,29 @@ def initialize(self, **kwargs):
...
         self.ti = 0  # The time index, e.g. 0, 1, 2
+        self.dt_year = ss.time_ratio(pars.unit, pars.dt, 'year', 1.0) # Figure out what dt is in years; used for demographics # TODO: handle None
```
RomeshA, line 162: It doesn't feel like we should need this - or otherwise have it here (if the conversion isn't automatic, I would have thought this would get handled in the demographic modules rather than at the sim level)

RomeshA (commented):
This is looking great! Feels like it's on the right track. I think the `TimePar` class might be too tightly connected to the parent, some of those conversions I was expecting to be automatic based on operators of the `TimePar` class rather than separate functions like `time_ratio`. For something like `ss.days(7)+ss.years(1)` I would have thought this would just directly return the equivalent of `ss.days(372)` (could debate whether it should return `ss.years(1+7/365)` but actually it _shouldn't_ matter because once the quantity has been defined, all conversions would be automatic). Left some other comments but my overall question is whether it's possible to remove the parent from `TimePar`.

From a workshop standpoint some of those changes probably don't matter since the important thing is that the interface in the modules is about right, and that part is looking quite good

starsim/disease.py
cliffckerr: Restored, was breaking STIsim

starsim/timeunits.py
```diff
@@ -0,0 +1,279 @@
...
+            self.set_x()
+        return self
+    
+    def set_factor(self):
+        """ Set factor used to multiply the value to get the output """
+        self.factor = time_ratio(unit1=self.unit, dt1=1.0, unit2=self.parent_unit, dt2=self.parent_dt)
+        return
+    
+    # Act like a float
+    def __add__(self, other): return self.x + other
```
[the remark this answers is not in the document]
cliffckerr: I think this could be possible, the question would just be (a) if there are performance considerations, (b) if there are corner cases where it wouldn't work. An issue is something like `beta`: for most diseases it has units of per-time, but for STIs it has units of per-act. So for most diseases it would make sense to require `number = beta*ss.dur` to remove dimensions, but this isn't _always_ the case.

tests/test_time.py
```diff
@@ -0,0 +1,139 @@
...
+    
+    # Test duration dt
+    d1 = ss.dur(2)
+    d2 = ss.dur(3)
+    d3 = ss.dur(2, parent_dt=0.1)
+    d4 = ss.dur(3, parent_dt=0.2)
+    for d in [d1,d2,d3,d4]: d.init()
+    
+    assert d1 + d2 == 2+3
+    assert d3 + d4 == 2/0.1 + 3/0.2
```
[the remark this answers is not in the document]
cliffckerr: Agree -- I think we're most of the way having `TimePar` be a wrapper rather than a thing-being-wrapped, e.g. `ss.dur(value=np.array([...])`. Then that would make it easier to keep the units, like how operations on a `pd.Series` usually return a `Series`.

starsim/diseases/syphilis.py
cliffckerr: True. I'm thinking we should have `tvec` (in module units) and `abs_tvec` (in sim units, starting at 0), and `dt` (in module units) and `abs_dt` (in sim units).

starsim/sim.py
cliffckerr: I'm always happy to make variable names shorter :D

starsim/sim.py
cliffckerr: It's the "shared source of truth" between the sim and the modules. `timearray` always starts at 0, is a float, and has sim time units. Whereas `timevec` can have different units, be a `datetime` array, etc.

starsim/sim.py
cliffckerr: That's true. I put it here since it's used both in the demographics modules and also in `People` for doing aging -- which is, perhaps confusingly, not part of demographics!

starsim/people.py
daniel-klein: Is `ss.rate(1)` technically correct here, as an alternative to `dt_year`?

cliffckerr: No, because there's no guarantee that any of the units are years. The sim and module could both have units of day, in which case `ss.rate(1)` is 1 day, not 1 year.

starsim/time.py
cliffckerr: We do, alas!

cliffckerr:
See https://github.com/starsimhub/starsim/issues/650 for outstanding issues

tests/test_time.py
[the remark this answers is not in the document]
cliffckerr: Done, this is working now.

PR #639, opened 2024-09-24
Distribution auto-jumping bug fix

daniel-klein:
Distributions keep track of the number of jumps in a variable named `ind`. With auto-jumping, `ind` is increased by 1 each time the distribution is called. At the beginning of each time step, we were calling jump() with `to=self.ti+1`, which sets all distribution `ind` values to ti+1.

The major problem here is that with auto-jumping, a user could call a distribution multiple times per time step. The jumps counter (`ind`) might start at say 1, and subsequently increase to 2, 3, and 4 during a step (if a user calls `rvs` several times). But then at the next time step, step() was calling `jump()` setting `ind` to 2, causing repeated random numbers to be used.

The easy one-line fix is to remove the `to=` argument to `jump` inside the `step` loop.

**NOTE:** This science bug fix could change simulation results.


cliffckerr:
Let's discuss in the meeting tonight -- this solves the bug you describe, but introduces a new bug such that if there is ever a difference in the number of RNG calls, the numbers for that dist won't match for the rest of the sim. I think the solution is to increment +1 per call, but increment by a large number (larger than the possible number of calls) per timestep.

cliffckerr:
Should be fixed by #638

cliffckerr:
Going to include this after all as v1.0.2.

cliffckerr (approved):
Implemented *1000 rather than None, but same idea!

PR #647, opened 2024-09-25
Make jump robust

cliffckerr:
Makes the changes to `jump` discussed on the call -- which already caught two bugs in the code (both of which were due to ancient holdovers from the `MultiRNG` days).

If these changes look good, I'll backport `jump_dt()` and the updated `jump()` to Starsim v1.0.

### DIFF
--- a/starsim/diseases/syphilis.py
+++ b/starsim/diseases/syphilis.py
@@ -210,10 +210,6 @@ class Syphilis(ss.Infection):
             self.set_secondary_prognoses(secondary_from_primary)
             self.rel_trans[secondary_from_primary] = self.pars.rel_trans['secondary']
 
-        # Hack to reset the MultiRNGs in set_secondary_prognoses so that they can be called again in this timestep. TODO: Refactor
-        self.pars.p_latent_temp.jump(ti+1)
-        self.pars.dur_secondary.jump(ti+1)
-
         # Secondary reactivation from latent
         secondary_from_latent = (self.latent_temp & (self.ti_secondary <= ti)).uids
         if len(secondary_from_latent) > 0:
--- a/starsim/distributions.py
+++ b/starsim/distributions.py
@@ -207,7 +207,7 @@ class Dist:
         self._slots = None # Internal variable to track currently-in-use slots
         
         # History and random state
-        self.dt_jump_size = 1000 # How much to advance the RNG for each timestep
+        self.dt_jump_size = 1000 # How much to advance the RNG for each timestep (must be larger than the number times dist is called per timestep)
         self.rng = None # The actual RNG generator for generating random numbers
         self.trace = None # The path of this object within the parent
         self.ind = 0 # The index of the RNG (usually updated on each timestep)
--- a/starsim/distributions.py
+++ b/starsim/distributions.py
@@ -109,7 +109,7 @@ class Dists(sc.prettyobj):
                 checked[seed] = dist
         return
 
-    def jump(self, to=None, delta=1):
+    def jump(self, to=None, delta=1, force=False):
         """ Advance all RNGs, e.g. to call "to", by jumping """
         out = sc.autolist()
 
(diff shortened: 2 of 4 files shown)


starsim/diseases/syphilis.py
```diff
@@ -210,10 +210,6 @@ def step_state(self):
             self.set_secondary_prognoses(secondary_from_primary)
             self.rel_trans[secondary_from_primary] = self.pars.rel_trans['secondary']
 
-        # Hack to reset the MultiRNGs in set_secondary_prognoses so that they can be called again in this timestep. TODO: Refactor
-        self.pars.p_latent_temp.jump(ti+1)
-        self.pars.dur_secondary.jump(ti+1)
```
daniel-klein, line 215: Nice to see these go away!

starsim/distributions.py
```diff
@@ -202,6 +207,7 @@ def __init__(self, dist=None, distname=None, name=None, seed=None, offset=None,
         self._slots = None # Internal variable to track currently-in-use slots
         
         # History and random state
+        self.dt_jump_size = 1000 # How much to advance the RNG for each timestep (must be larger than the number times dist is called per timestep)
```
daniel-klein, line 210: Thoughts on making the dt jump size a parameter? I know jumping is fast, but on ti=10000 is there a performance implication of the 1000x increase in jump size? For simulations implemented to call each distribution only once per time step, this could be set to 1. But I also see value in this being an intrinsic, and wouldn't want people to INCREASE it due to misunderstanding of distributions.

starsim/distributions.py
```diff
@@ -324,26 +330,39 @@ def reset(self, state=0):
         self.ready = True
         return self.state
 
-    def jump(self, to=None, delta=1):
+    def jump(self, to=None, delta=1, force=False):
         """ Advance the RNG, e.g. to timestep "to", by jumping """
         
-        # Do not jump if centralized
+        # Do not jump if centralized # TODO: remove
```
daniel-klein, line 336: I still use this, and it could improve performance in some settings where CRN is not a consideration.

starsim/distributions.py
```diff
@@ -324,26 +330,39 @@ def reset(self, state=0):
...
+    def jump_dt(self, ti=None, force=False):
+        """
+        Automatically jump on the next value of dt
+        
+        Args:
+            ti (int): if specified, jump to this timestep (default: current sim timestep)
+        """
+        if ti is None:
+            ti = self.sim.ti
+        to = self.dt_jump_size*(self.sim.ti+1)
```
daniel-klein, line 364: BUG: This should be `ti` not `self.sim.ti`

daniel-klein: How did this get through testing? Perhaps add a test to catch this bug?

daniel-klein (commented):
Just fix the one bug I found and consider adding a test

starsim/distributions.py
cliffckerr: There's zero performance impact, so think this is ok:
```py
import sciris as sc
import starsim as ss

reps = 1e4
deltas = [1, 10, 1_000, 10_000, 100_000]
for delta in deltas:
    d = ss.random(strict=False)
    with sc.timer(f'reps={reps:n} | delta={delta:n}'):
        for r in range(int(reps)):
            d.jump(delta=delta)

reps=10,000 | delta=1: 0.247 s
reps=10,000 | delta=10: 0.353 s
reps=10,000 | delta=1,000: 0.267 s
reps=10,000 | delta=10,000: 0.265 s
reps=10,000 | delta=100,000: 0.198 s
```

Jumping is slow, though! Takes 50% as much time as generating the random numbers

starsim/distributions.py
cliffckerr: If we're going to keep it in longer term we should come up with a better name for it than `_centralized`, though!

starsim/distributions.py
cliffckerr: Good catch, fixed!

PR #655, opened 2024-09-27
Improvements from "spokes"

daniel-klein:
* Revised calibration interface, see test_calibration.py for example
* Agents now have a `parent` which indicates the uid of the parent if using Pregnancy and nan otherwise. Could put a sentinel value here for seed or import cases, but haven't done-so yet.
* Pregnant updates:
  - women now track `child_uid`
  - On neonatal death, the pregnancy state of the mother is corrected
  - Pregnancy rates now adjusted for infecund rather than pregnant
  - Pregnancy now has a burnin, which defaults to True
  - Pregnancy gains `p_neonataldeath_on_maternaldeath` to capture death of the unborn neonate if the mother dies
* Slots now has a minimum, default of 100, to account for small initial population sizes that grow dramatically over time
* cmr and cbr results are now floats
*  Added a gamma distribution


daniel-klein:
One thing I wasn't sure about was setting the age of the new agents in Pregnancy, [here](https://github.com/starsimhub/starsim/blob/d115622d1bdb61259536ca0c5750b262070e2e23/starsim/demographics.py#L479)

starsim/demographics.py
```diff
@@ -251,8 +251,8 @@ def __init__(self, pars=None, metadata=None, **kwargs):
             dur_postpartum = ss.lognorm_ex(mean=ss.years(0.5), std=ss.years(0.5)), # Duration for post-natal transmission (e.g. via breastfeeding)
             fertility_rate = 0, # Can be a number of Pandas DataFrame
             rel_fertility = 1,
-            maternal_death_prob = ss.bernoulli(0),
-            p_neonataldeath_on_maternaldeath = ss.bernoulli(0),
+            p_maternal_death = ss.bernoulli(0),
+            p_neonatal_death = ss.bernoulli(0),
```
daniel-klein, line 255: Do you think this variable name is sufficient? This is not the probability of neonatal death (overall), but rather the probability of neonatal death on the event of the death of the mother during pregnancy.

cliffckerr: Thinking about it more, it should actually be called `p_fetal_death`, right? If the mother is still pregnant, the baby hasn't been born, so is a fetus rather than a neonate.

cliffckerr:
Added plotting to calibration, including data plotting to the sim!
image

starsim/demographics.py
cliffckerr: Ah, sorry, I was confused. I didn't realize that fetal deaths were handled separately, and didn't see that `child_uids` got reset after the postpartum period. Yes, I think the name is ok given that we're in the `Pregnancy` module.

cliffckerr: Wait no, still confused. Your comment above refers the "unborn neonate" (i.e. fetus), but in the code it looks like this is handling both fetal and neonatal agents (up to the end of the postpartum period).

Issue #696, opened 2024-10-12
NCD does not run without log=True


daniel-klein:

daniel-klein:
Issue addressed in #697

Issue #692, opened 2024-10-08
Can't use a dist in more than one state anymore  - 'define_states'


menriquez-IDM:
(Context: I am porting tbsim to starsim 2.0 ).... in our Malnutrition Disease Module, as part of the constructor we are using distributions as default values, which used to work fine, but now are failing:

```python
    def __init__(self, pars=None, **kwargs):
        super().__init__(**kwargs)
        self.define_pars(
            beta = 1.0,         # Transmission rate  
            init_prev = 0.001,  # Initial prevalence 
        )
        self.update_pars(pars, **kwargs)

        anthro_path = os.path.join(DATADIR, 'anthropometry.csv')
        self.LMS_data = pd.read_csv(anthro_path).set_index('Sex')

        # Adding Malnutrition states to handle the Individual Properties related to this disease 
        self.define_states(
            # Hooks to the RATIONS trial
            ss.BoolArr('receiving_macro', default=False), # Determines weight trend
            ss.BoolArr('receiving_micro', default=False), # Determines micro trend

            "----->>>>>>>>>>>>>>>>>>>>>> ISSUE TRIGGERED BY DEFAULT VALUES BELOW:
            ss.FloatArr('height_percentile', default=ss.uniform()), # Percentile, stays fixed
            ss.FloatArr('weight_percentile', default=ss.uniform()), # Percentile, increases when receiving micro, then declines?
            ss.FloatArr('micro', default=ss.uniform()), # Continuous? Normal distribution around zero. Z-score, sigmoid thing. Half-life.


        )
        self.dweight = ss.normal(loc=self.dweight_loc, scale=self.dweight_scale)
        return
```

The error is as follows:
```bat
(venvVer2) c:\git\tbsim\scripts\general>python run_malnutrition.py
Starsim 2.0.0 (2024-10-01) — © 2023-2024 by IDM
Initializing sim with 10000 agents
Traceback (most recent call last):
  File "c:\git\tbsim\scripts\general\run_malnutrition.py", line 32, in <module>
    sim_n.run()
  File "c:\git\tbsim\venvVer2\Lib\site-packages\starsim\sim.py", line 298, in run
    if not self.initialized: self.init()
                             ^^^^^^^^^^^
  File "c:\git\tbsim\venvVer2\Lib\site-packages\starsim\sim.py", line 161, in init
    self.init_dists() # Initialize distributions
    ^^^^^^^^^^^^^^^^^
[...]

PR #697, opened 2024-10-12
Adds mixing pools

daniel-klein:
### Description
Add a single `MixingPool` and multiple `MixingPools` transmission routes to Starsim.

I like the interface, see the revised `tut_transmission.ipynb` or `test_mixingpools.py` for usage examples.

PR also addresses #692 and #696

Discussion points:
* Mixing pools are implemented as interventions rather than a network. Pools are NOT Networks (that's the point), and don't have `p1`, `p2`, or `beta`. Eventually, it would be nice to refactor `Networks` --> `Transmission` and have both Networks and MixingPools be subclasses of transmission that could handled similarly. That, however, would likely be a breaking change.
* Performance doesn't seem great, `RandomNet` is faster in the little bit of exploration that I did. Code could benefit from jitting. 
* Should add more docstrings.

### Checklist


daniel-klein:
Also now passing **kwargs through Module initialization, which seemed like an oversight.

PR #727, opened 2024-10-30
Mixing pools ck

cliffckerr:
Refactor of mixing pools branch prior to merging with `rc2.1.0`.
- Clears the output of `tut_transmission.ipynb` (should be cleared prior to committing)
- Fixes `timepar.to()` to not do double conversion
- Moves mixing pools from `interventions.py` to `networks.py`
- Adds some documentation to mixing pools
- Minor refactoring of mixing pools
- Improves timepar `repr`
- Changes `ss.days()` etc. from classes to functions
- Adds `ss.rate_prob()`
- Fixes tests


starsim/distributions.py
```diff
@@ -11,19 +11,15 @@
...
     Cannot use Python's built-in hash() since it's randomized for strings. Hashlib
     is 5x slower than int.from_bytes(string.encode(), byteorder='big'), but should
     only add a couple milliseconds to a typical sim.
     """
-    integer = int.from_bytes(string.encode(), byteorder='big')
-    # integer = sc.sha(string, asint=True) # Hash the string to an integer
-    if modulo is None:
-        return integer
-
+    integer = sc.sha(string, asint=True) # Hash the string to an integer
```
daniel-klein, line 22: Consider updating the comment above to mention sc.sha

daniel-klein: How did you choose 10M as the modulo?

starsim/networks.py
```diff
@@ -987,3 +994,191 @@ class PostnatalNet(MaternalNet):
...
+            self.ti_cache = sim.ti
+        return self.uids
+
+    def __repr__(self):
+        return f'age({self.low}-{self.high})'
+
+
+class MixingPools(Route):
+    """
+    A container for one or more MixingPool instances
```
daniel-klein, line 1029: Kinda, but not quite. It's not so much a "container" as it is a builder to create a rectangular matrix of MixingPool instances.

starsim/networks.py
```diff
@@ -987,3 +994,191 @@ class PostnatalNet(MaternalNet):
...
+
+    By default, separates the population into <15 and >15 age groups.
+    """
+    def __init__(self, pars=None, **kwargs):
+        super().__init__(**kwargs)
+        self.define_pars(
+            diseases = None,
+            beta = ss.beta(0.1),
+            contact_matrix = np.array([[2.4, 0.49], [0.91, 0.16]]), # TODO: document where these came from
```
daniel-klein, line 1038: There is nothing to document here, the values are made up and intended to be replaced by the user - as are the `src` and `dst`. Is there a better way, perhaps taking `None` as the default and throwing an error if user input is not supplied?

starsim/networks.py
```diff
@@ -987,3 +994,191 @@ class PostnatalNet(MaternalNet):
...
+            acq = self.eff_contacts[self.dst_uids] * disease.susceptible[self.dst_uids] * disease.rel_sus[self.dst_uids]
+            p = self.pars.beta * trans * acq #1 - np.exp(-self.pars.beta * trans * acq)
+
+            self.p_acquire.set(p=p)
+            new_cases = self.p_acquire.filter(self.dst_uids)
+            n_new_cases += len(new_cases)
+
+            disease.set_prognoses(new_cases)
+
+        return n_new_cases
```
daniel-klein, line 1184: Missing new line at end of file.

daniel-klein (commented):
A few recommended change, including `rate_prob`

tests/benchmark.json
```diff
@@ -1,12 +1,12 @@
 {
   "time": {
-    "initialize": 0.034,
-    "run": 0.772
+    "initialize": 0.036,
+    "run": 0.957
```
daniel-klein, line 4: Higher CPU performance and longer run time - should we be worried?

starsim/networks.py
```diff
@@ -987,3 +994,191 @@ class PostnatalNet(MaternalNet):
...
+    def get_uids(self, func_or_array):
+        if func_or_array is None:
+            return self.sim.people.auids
+        elif callable(func_or_array):
+            return func_or_array(self.sim)
+        elif isinstance(func_or_array, ss.uids):
+            return func_or_array
+        raise Exception('src must be either a callable function, e.g. lambda sim: ss.uids(sim.people.age<5), or an array of uids.')
+
+    def remove_uids(self, uids):
```
daniel-klein, line 1149: If using `MixingPool` as an intervention in a Sim, this code will not be called, right?

daniel-klein (changes requested):
Some changes required for the `rate_prob`

tests/benchmark.json
cliffckerr: No, "run" is calculated as actual runtime multiplied by "cpu_performance" (i.e., your run probably took about 1 second). In other words, "run" should be consistent regardless of "cpu_performance". I don't know why for you, both were so much lower -- I think the "cpu_performance" calculation wasn't quite accurate.

starsim/demographics.py
```diff
@@ -402,7 +402,7 @@ def do_step(self):
         """ Perform all updates """
         self.update_states()
         conceive_uids = self.make_pregnancies()
-        self.n_pregnancies += len(conceive_uids)
+        self.n_pregnancies = len(conceive_uids)
```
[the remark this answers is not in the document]
cliffckerr: Ah ok, that seems like it might be an anti-pattern, but will revert for now!

starsim/distributions.py
cliffckerr: Updated. 10M is intended to be a compromise between human readability and absurdly low probability of collisions. But actually the probability isn't as low as I thought, so I upped this to 1 billion (which is now an average of roughly one collision per million sims -- certainly not enough to affect results!).

starsim/networks.py
cliffckerr: I've removed the defaults and added them to a docstring example instead.

starsim/networks.py
```diff
@@ -987,3 +994,191 @@ class PostnatalNet(MaternalNet):
...
+        dst (inds): destination agents; as above
+        beta (float): overall transmission
+        contacts (Dist): the number of effective contacts of the destination agents
+    """
+    def __init__(self, pars=None, **kwargs):
+        super().__init__(**kwargs)
+
+        self.define_pars(
+            diseases = None,
+            src = None,
```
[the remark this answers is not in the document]
cliffckerr: Why not assume that `src=None` is all agents?

starsim/networks.py
cliffckerr: Correct, which will lead to bugs in the corner case of running with `src` or `dst` being explicit UIDs with demographics turned on.

starsim/networks.py
cliffckerr: Added

starsim/timepars.py
```diff
@@ -365,6 +390,34 @@ def update_values(self):
         return self.values
 
 
+class rate_prob(TimePar):
```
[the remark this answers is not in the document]
cliffckerr: Well spotted, fixed!

Issue #776, opened 2024-11-18
Refactor calibration


cliffckerr:
- [ ] Ensure `ss.Calibration().calibrate()` works as easily as possible
- [ ] Simplify `build_sim`
- [ ] Update API to ensure consistency
- [ ] Improve plotting (e.g. include data)
- [ ] Simplify `suggest_type`
- [ ] Reconsider names for `beta`/`beta_binomial` etc
- [ ] Improve docstrings
- [ ] Consider weighted interval score / energy scores

Consider moving this to `check_fit()`:
```py
    if calib.after_fits.mean() <= calib.before_fits.mean():
        print('✓ Calibration improved fit')
    else:
        print('✗ Calibration did not improve fit, but this sometimes happens stochastically and is not necessarily an error')
```

cliffckerr:
Closed by #865

PR #783, opened 2024-11-21
Tiny fixes in test calib and calibration

pausz:
### Description
- In `test_calibration.py`:
    - Super tiny fix: moves line that could  have resulted in an error or a mistake for 'rand_seed', depending on the order of keys in calib parameters.
    - Fix path to parameters in `calib_pars`
- In `calibration.py`: 
    - Remove `rand_seed` from `calib_pars` before self.build_fn is called. `rand_seed`'s value is an int not a dict like for the rest of the calib parameters
 
image

### DIFF
--- a/starsim/calibration.py
+++ b/starsim/calibration.py
@@ -88,6 +88,9 @@ class Calibration(sc.prettyobj):
         sim = sc.dcp(self.sim)
         if label: sim.label = label
 
+        if 'rand_seed' in calib_pars:
+            sim.pars['rand_seed'] = calib_pars.pop('rand_seed')
+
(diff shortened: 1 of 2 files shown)


robynstuart:
Added one more little change here to [this line](https://github.com/starsimhub/starsim/blob/main/starsim/calibration.py#L466), pretty sure this is a typo

cliffckerr (approved):
Thanks! I think we'll be doing a pretty big revamp of the calibration interface (see https://github.com/starsimhub/starsim/issues/776), but these are helpful fixes in the interim!

Issue #800, opened 2024-12-04
Bug in identifying parent for units


daniel-klein:
Result is that many TimePars think the parent units are 'year' (from demographics), resulting in wrong rates everywhere.

robynstuart:
I think this is the issue behind https://github.com/starsimhub/starsim/issues/794

cliffckerr:
Closed by #831

PR #801, opened 2024-12-04
Addressing Bug in identifying parent for units #800

daniel-klein:
### Description
Addressing a bad bug in units

### Checklist


cliffckerr:
Included in #810

PR #810, opened 2024-12-07
[DRAFT] Improved calibration

daniel-klein:
### Description
Still need to update the calibration tutorial, should be easy based on `test_calibration.py`.

### Checklist


starsim/calib_components.py
```diff
@@ -0,0 +1,379 @@
...
+    A class to compare a single channel of observed data with output from a
+    simulation. The Calibration class can use several CalibComponent objects to
+    form an overall understanding of how will a given simulation reflects
+    observed data.
+
+    Args:
+        name (str) : the name of this component. Importantly, if
+            extract_fn is None, the code will attempt to use the name, like
+            "hiv.prevalence" to automatically extract data from the simulation.
+        expected (df) : pandas DataFrame containing calibration data. The index should be the time 't' in either floating point years or datetime.
```
pausz, line 23: Perhaps reword as `pandas DataFrame containing the reference data to calibrate to.`

> The index should be the time 't' in either floating point years or datetime.

1. What would be the recommendation if the reference data have been aggregated over multiple years? Some of the calibration targets in typhoid have the concept of "YearBin", meaning the reference data has been aggregated over multiple years.  As a workaround I have set the time index as the midpoint between year_bin_start and year_bin_end.

2. Could the index ever be something other than time? `expected` could be more generally a pandas Series with an index that must be monotonic. 
     Drawing again from the calibration targets we have in typhoid, one of them is a proportion of cases, but the data have been aggregated (sum) over 4 years, and stratified by age. The reference data consists of 4 data points (1 timepoint x 4 data points of 4 age groups).  With the current implementation we have to define 4 CalibComponents, which seems excessive -- whether calibrating to such a reduced target is a good idea deserves a separate discussion :sweat_smile:. If `expected` could be indexed by something other than time, it would be great. 
     
     Thinking out loud, it is already possible to pass age as the index, since `t` can be a float, but doing that feels hacky.

starsim/calib_components.py
```diff
@@ -0,0 +1,379 @@
...
+        extract_fn (callable) : A callable function used to extract data from a sim object, which is passed as the lone argument - should look like expected.
```
pausz, line 24:
```python
"""
        extract_fn (callable) : a function to extract predicted/actual data in the same
            format and with the same columns as `expected`.
"""
```

starsim/calib_components.py
```diff
@@ -0,0 +1,379 @@
...
+        conform (str/callable): To handle misaligned timepoints between observed data and simulation output, it's important to know if the data are incident (like new cases) or prevalent (like the number infected).
+            If 'prevalent', simulation outputs will be interpolated to observed timepoints using 't'.
+            If 'incident', outputs will be computed as differences between cumulative values at 't' and 't1'.
```
pausz, line 27: When I first read an earlier version of this I had a hard time understanding how conform actually did, just from reading the description.  My docstrings for this argument after seeing the code:

``` python
"""
        conform (str | callable): specify how to handle timepoints that don't
            align exactly between the expected and the actual/predicted/simulated
            data so they conform to a common time grid. Whether the data represents
            a 'prevalent' or an 'incident' quantity impacts how this alignment is performed.

            If 'prevalent', it means data in expected & actual dataframes represent
            the current state of the system, like the number of currently infected
            individuals. In this case, the data in 'simulated' or 'actual'  will be interpolated
            to match the timepoints in 'expected', allowing for pointwise comparisons
            between the expected and actual data.
"""
``` 
Is  `t1` the next timpoint and should exist as an index in `expected`?

starsim/calib_components.py
```diff
@@ -0,0 +1,379 @@
...
+                raise ValueError(errormsg)
+        else:
+            if not callable(conform):
+                errormsg = f'The conform argument must be a string or a callable function, not {type(conform)}.'
+                raise TypeError(errormsg)
+            self.conform = conform
+
+        return
+
+    @staticmethod
```
pausz, line 55: Any reason not to move them outside the class? This function can be useful on their own for applications other than calibration.

starsim/calib_components.py
```diff
@@ -0,0 +1,379 @@
...
+        kwargs: Additional arguments to pass to the likelihood function
+    """
+    def __init__(self, name, expected, extract_fn, conform, weight=1, include_fn=None):
+        self.name = name
+        self.expected = expected
+        self.extract_fn = extract_fn
+        self.weight = weight
+        self.include_fn = include_fn
+
+        if isinstance(conform, str):
```
pausz, line 39:
```python
        # __init__(....)
        self.include_fn = include_fn
        self.avail_conforms = {"incident":  linear_accum,  # or self.linear_accum if left as staticmethod
                               "prevalent": linear_interp}

        self.conform = self._validate_conform(conform)
        
        return 

    def _validate_conform(self, conform):
        # Fail early
        if not isinstance(conform, str) and not callable(conform):
            raise Exception(f"The conform argument must be a string or a callable function, not {type(conform)}.")
        elif isinstance(conform, str):
            conform_ = self.avail_conforms.get(conform)
            if conform_ is None:
                avail = self.avail_conforms.keys()
                raise ValueError(f"The conform argument must be one of {avail}, not {conform}.")
        else:
            conform_ = conform
        return conform_
````

File: starsim/calib_components.py [hunk 6]
```diff
@@ -0,0 +1,379 @@
...
+
+    @staticmethod
+    def linear_accum(expected, actual):
+        """
+        Interpolate in the accumulation, then difference.
+        Use for incident data like incidence or new_deaths
+        """
+        t0 = np.array([sc.datetoyear(t) for t in expected.index.get_level_values('t').date])
+        t1 = np.array([sc.datetoyear(t) for t in expected.index.get_level_values('t1').date])
+        sim_t = np.array([sc.datetoyear(t) for t in actual.index.date if isinstance(t, dt.date)])
```
pausz, line 76: `actual_t`?

starsim/calib_components.py
```diff
@@ -0,0 +1,379 @@
...
+        return conformed
+
+    @staticmethod
+    def linear_accum(expected, actual):
+        """
+        Interpolate in the accumulation, then difference.
+        Use for incident data like incidence or new_deaths
+        """
+        t0 = np.array([sc.datetoyear(t) for t in expected.index.get_level_values('t').date])
+        t1 = np.array([sc.datetoyear(t) for t in expected.index.get_level_values('t1').date])
```
pausz, line 75: I added an assertion for the index in `expected`. If it fails, it means we can't interpolate the cumulative quantity. I ran into a case where the expected data had time in datetime, and the sampling period was monthly (2021-01-01, 2021-02-01, 2021-03-01, 2021-04-01, etc). When converted to a float representation, these timepoints are not regularly spaced.  

Btw, this was before `t1`
```python
    common_time_grid = expected.index
    expected_sampling_period = np.diff(common_time_grid )
    assert np.all(expected_sampling_period == expected_sampling_period[0])  # Check we have regularly sampled data

    # Make time grid for cumulative
    cum_time_grid = np.append(common_time_grid, common_time_grid[-1] + expected_sampling_period[0])  # Add one more because later we'll diff
  ```

File: starsim/calibration.py [hunk 8]
```diff
@@ -371,203 +358,61 @@ def to_json(self, filename=None, indent=2, **kwargs):
...
-    def __call__(self, sim):
-        return self.eval(sim)
-
-    def __repr__(self):
-        return f'Calibration component with name {self.name}'
-
-    def plot(self):
-        NotImplementedError
\ No newline at end of file
+    def plot_optuna(self, methods=None):
```
pausz, line 392: :heart_hands:

starsim/calib_components.py
daniel-klein: Good points, @pausz.
1. There are currently only two classes of conformers. The prevalent variety looks at a point in time (stock) whereas the incident variety looks at an accumulation (flow). If the Typhoid data were collected over a period of time, would it be more appropriate to use the incident conformer? Otherwise, we could think about creating some other conformers, would you want the average over the YearBins?
2. When I had initially started this design, I did imagine that the index could be anything (not just time). But as the implementation unfolded, I realized that time _always_ matters and is different from other categorical variables like age in how conformation works. But it does seem like an improvement would be to allow additional "axes" like age and sex. I'll create an issue.

starsim/calib_components.py
daniel-klein: Thanks

starsim/calib_components.py
daniel-klein: Thank you!

starsim/calib_components.py
daniel-klein: I like it, thanks!

starsim/calib_components.py
daniel-klein: Thanks for the suggestion, moved.

starsim/calibration.py
daniel-klein: :)

starsim/calib_components.py
daniel-klein: meh, it's a temp loop var

starsim/calib_components.py
daniel-klein: Could you please say more, @pausz? I don't see the issue with irregularly space time points considering we interpolate at both t and t1 below.

PR #831, opened 2024-12-20
Version 2.3

cliffckerr:
### Description
Updates calibration and mixing pools, and fixes a time parent bug.

### Checklist


starsim/sim.py
```diff
@@ -418,37 +418,40 @@ def shrink(self, inplace=True, size_limit=1.0):
...
+                mod.sim = shrunk
+                for state in mod.states:
+                    with sc.tryexcept():
+                        state.people = shrunk
+                        state.raw = shrunk
+
+                # Check that the module successfully shrunk
+            if size_limit:
+                for mod in sim.modules:
+                    size = sc.checkmem(mod, descend=0).bytesize[0]/1e6
```
daniel-klein, line 451: Check that this is total, and not just the first item.

starsim/sim.py
```diff
@@ -418,37 +418,40 @@ def shrink(self, inplace=True, size_limit=1.0):
...
-                if size > size_limit:
-                    warnmsg = f'Module {mod.name} did not successfully shrink: {size:0.1f} MB > {size_limit:0.1f} MB'
-                    ss.warn(warnmsg)
+                mod.sim = shrunk
+                for state in mod.states:
+                    with sc.tryexcept():
+                        state.people = shrunk
+                        state.raw = shrunk
+
+                # Check that the module successfully shrunk
```
daniel-klein, line 448: Shift comment <<

starsim/sim.py
daniel-klein: I'm still getting the did not successfully think message for just about every module in every simulation, creating LOTS of verbose output when running thousands of sims! Size is 1.1 MB, just above the 1MB threshold. Consider increasing.

PR #843, opened 2025-01-17
Operators work directly on raw arrays if most agents are active

RomeshA:
### Description

Operators on `Arr` and `BoolArr` use `asnew()` to return new `Arr` instances. These work along the lines of

- Using `self.values` to create a numpy array with length equal to `auids`
- Creating a corresponding array from `other.values` if `other` is an `Arr`
- Creating an intermediate result array with length equal to `auids`
- Creating a new empty array the same size as `uids`
- Populating it sparsely with the result array based on `auids` 

The intention is that the alternative of operating on every UID (i.e., on the `raw` arrays directly) would be unnecessarily slow if there are many inactive agents. However, in practice the overhead associated with creating the intermediate arrays and then sparsely inserting it is fairly significant. For simulations that do not have a significant proportion of inactive agents, it can be faster to just operate on every entry. In that case, no intermediate arrays need to be created - the output of the logical operation of the two raw arrays can be directly used as the raw array for the new `Arr` instances. In some simulations this can give a speedup of around 40%.

This PR thus adds an attribute for `Arr` that corresponds to whether this optimization should be used, with a threshold of 50% (i.e., it will use the raw entries if over half the agents are active). This threshold could be tuned as we get more use cases. There is a corresponding update to `asnew` to optionally take in a `raw` array, which will be used if provided.

There should be no changes necessary in any user code and simulation output should be the same

### Checklist

### DIFF
--- a/starsim/arrays.py
+++ b/starsim/arrays.py
@@ -233,12 +233,47 @@ class Arr(BaseArr):
         self.raw[key] = value
         return
 
-    def __gt__(self, other): return self.asnew(self.values > other,  cls=BoolArr)
-    def __lt__(self, other): return self.asnew(self.values < other,  cls=BoolArr)
-    def __ge__(self, other): return self.asnew(self.values >= other, cls=BoolArr)
-    def __le__(self, other): return self.asnew(self.values <= other, cls=BoolArr)
-    def __eq__(self, other): return self.asnew(self.values == other, cls=BoolArr)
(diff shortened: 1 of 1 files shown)


RomeshA:
Going to mark this as draft as other changes in our code seem to have made this change have less impact, will continue profiling further

cliffckerr:
Closing since should be largely superseded by new indexing system, including Numba indexer

Issue #846, opened 2025-01-24
Remove ss.beta


daniel-klein:
Transmission for networks or mixing pools is a per-agent probability computed as rel_sus * rel_trans * beta. That product should be converted from a per-time rate to a probability. It makes no sense for just `beta` to be a probability. Recommend that we do away with `ss.beta` and use rates (ss.perday, ss.peryear) instead. The unit- and dt- conversions will need to be handled after computing the product.

In TBsim, rel_sus is small for many agents, so we need beta > 1 to achieve desired levels of transmission.

RomeshA:
We had a related question come up when using mixing pools as to how `beta` should be specified and in particular, what time conversions it should be subject to. I think it would be great for us to be clearer about how the probability of infection is broken up into a contact rate (over time) and a probability of transmission per contact. That differentiation could be useful as well when considering co-circulating diseases on the same network e.g., with STIs, where the contact rate for different diseases must be identical but the transmission probability per contact might not be. In general then it would be the network contact rate that gets converted from a rate into a number depending on dt, and then the disease's transmission probability per contact would actually not be a time parameter at all

daniel-klein:
To further elevate this issue, I'll just note that a user cannot simply override a default `TimePar` parameter with a scalar. For example, in the SIR module, `beta` is a `ss.beta`. If a user try to set a floating point value for `beta`, say 2.0, `update_pars` calls `_update_timepar` which ultimately calls `.set` on the default `ss.beta` with the new value of 2.0, resulting in an updated `ss.beta` rather than a float. Of course, a `ss.beta` with value of 2.0 is not permitted, so execution stops.

Instead, a user would need substitute a different TimePar, like a `ss.rate_prob(2.0, unit='days')` to avoid the model stopping prematurely.

RomeshA:
Noting that #847 removes `ss.beta` and instead the user can set the module's beta to be a `Rate`, `TimeProb`, or `RateProb` depending on how they would like the value to scale with `dt` (it can also be left as a number which will show the same warning as on `main`)

RomeshA:
@robynstuart wonders about `ss.betaperday` https://github.com/starsimhub/starsim/pull/847#discussion_r2068127552 (if we are keeping `ss.beta`)

PR #847, opened 2025-01-28
Time experiments

RomeshA:
Work in progress branch for brainstorming ways to make working with durations/rates a bit more automatic and safer by reducing reliance on the parent dt.

API examples are currently in `starsim/time.py` (that file can be run directly as a demo)

Still in progress, the `Time` class with the time arrays and dists still need a little work before the models can be run. The current plan would be for `Dist` objects to have pars that retain their original units, and then to internally do conversion to their parent dt and cache the bare values for use with `rvs()` (so they would still return bare values for consumption by their parent module, for performance reasons. It seems like it would not be practical for `Dist` objects to return arrays of dur/rate/other objects). But other modules could check the dist object to access the parameter values with full duration information without having to also convert using the correct parent dt)


RomeshA:
@cliffckerr slight dependency on https://github.com/sciris/sciris/pull/642 which facilitates migration of application code that uses findnearest e.g., STIsim

RomeshA:
Example usage: https://github.com/starsimhub/starsim/blob/time-experiments/tests/devtests/Time%20API.ipynb
See also
- STIsim updates (all tests passing) - https://github.com/starsimhub/stisim/tree/time-experiments (branched off `jc-bv2`)
- Vagisim updates (most tests passing) - https://github.com/starsimhub/vagisim/tree/time-experiments (branched off `jc-dec10`)
- hiv_vmb_kenya (example of application migration - tested using `model.py` confirming model runs and results unchanged) - https://github.com/starsimhub/hiv_vmb_kenya/tree/time-experiments (branched off `hiv-additions`)

daniel-klein (commented):
One minor point is that we often integrate rates over time using an expoential, but the result is a bad operator without parenthesis around the rate*dt:

np.exp(-self.pars.waning * self.t.dt)

  File "<string>", line 1, in <module>
TypeError: bad operand type for unary -: 'Rate'

starsim/diseases/sir.py
```diff
@@ -133,7 +133,7 @@ def step_state(self):
 
     def update_immunity(self):
         has_imm = (self.immunity > 0).uids
-        self.immunity[has_imm] = (self.immunity[has_imm])*(1 - self.pars.waning)
+        self.immunity[has_imm] = (self.immunity[has_imm])*(1 - self.pars.waning*self.t.dt)
```
daniel-klein, line 136: It's fine to view waning as a geometric process, but the result will be slightly sensitive to dt. Instead, consider:
```python
factor = np.exp(-(self.pars.waning*self.t.dt))
self.immunity[has_imm] = (self.immunity[has_imm])*factor
```
The `factor` could be cached.

starsim/diseases/hiv.py
```diff
@@ -11,21 +11,21 @@
...
+            beta = ss.TimeProb(1.0), # Placeholder value
             cd4_min = 100,
             cd4_max = 500,
             cd4_rate = 5,
             eff_condoms = 0.7,
             art_efficacy = 0.96,
             init_prev = ss.bernoulli(p=0.05),
             death_dist = ss.bernoulli(p=self.death_prob_func), # Uses p_death by default, modulated by CD4
-            p_death = ss.rate(0.05), # NB: this is death per unit time, not death per infection
+            p_death = ss.peryear(0.05), # NB: this is death per unit time, not death per infection
```
daniel-klein, line 26: `peryear` is a rate, but we actually want a probably here. I think `RateProb` would be a better choice.

starsim/diseases/hiv.py
```diff
@@ -39,7 +39,7 @@ def __init__(self, pars=None, *args, **kwargs):
     @staticmethod
     def death_prob_func(module, sim, uids):
         p = module.pars
-        out = p.p_death / (p.cd4_min - p.cd4_max)**2 *  (module.cd4[uids] - p.cd4_max)**2
+        out = (p.p_death*module.t.dt) / (p.cd4_min - p.cd4_max)**2 *  (module.cd4[uids] - p.cd4_max)**2
```
daniel-klein, line 42: Oh, just realizing that users are once again required to multiply by dt. I think it's a good change, but we will need to communicate this clearly to end users considering their code will need to change.

starsim/time.py
```diff
@@ -227,70 +108,871 @@ def from_year(cls, year):
...
+            if self.value == 0:
+                return 0
+            else:
+                factor = self.period/other
+                return 1 - np.exp(-self.value/factor)
+        else:
+            return self.__class__(self.value*other, self.period)
+
+    def __truediv__(self, other): raise NotImplementedError()
+    def __rtruediv__(self, other): raise NotImplementedError()
```
daniel-klein, line 959: Avoiding NotImplementedError for straightforward rate multiplication.

daniel-klein (commented):
I worry that the unit-dependence in multiplication of RateProb and TimeProb will be confusing to many users. Relatedly, the net result is a strong dependence on the order of operations. Consider:
```python
p = ss.RateProb(0.9)

p * 2 * sim.t.dt
0.8347011117784134

p * sim.t.dt * 2.
1.1868606805188018
```

starsim/diseases/cholera.py
```diff
@@ -118,7 +118,7 @@ def calc_environmental_prev(self):
         old_prev = self.results.env_prev[ti-1]
 
         new_bacteria = p.shedding_rate * (n_symptomatic + p.asymp_trans * n_asymptomatic)
-        old_bacteria = old_prev * (1 - p.decay_rate)
+        old_bacteria = old_prev * (1 - p.decay_rate*self.t.dt)
```
daniel-klein, line 121: shedding_rate is a rate, and therefore needs to be multiplied by a Dur to get a value

daniel-klein, line 121: We probably want to switch to continuous compounding so that
```python
old_bacteria = np.exp(-p.decay_rate * self.t.dt) * old_prev
```
I made issue #925

starsim/time.py
```diff
@@ -227,70 +108,875 @@ def from_year(cls, year):
...
+    def __repr__(self):
+        return f'<{self.__class__.__name__}: {self.value} per {self.period.str()}>' # Use str to get the friendly representation
+
+    def __mul__(self, other):
+        if isinstance(other, np.ndarray):
+            return other*self
+        elif isinstance(other, Dur):
+            return self.value*other/self.period
+        elif sc.isnumber(other) and other == 0:
+            return 0 # Or should this be a rate?
```
daniel-klein, line 793: YES, should be a rate! Otherwise, rate * some_value * dt gives an answer that is unit-less most of the time, but has units of Dur if some_value is zero.

starsim/demographics.py
```diff
@@ -23,15 +23,14 @@ class Demographics(ss.Module):
 
 class Births(Demographics):
     """ Create births based on rates, rather than based on pregnancy """
-    def __init__(self, pars=None, metadata=None, **kwargs):
+    def __init__(self, metadata=None, **kwargs):
         super().__init__()
         self.define_pars(
-            unit = 'year',
             birth_rate = ss.peryear(20),
             rel_birth = 1,
```
daniel-klein, line 30: Changing from peryear to RateProb so that we get a probability (via exponential) after multiplication by dt.

starsim/demographics.py
```diff
@@ -86,12 +88,7 @@ def get_births(self):
...
             this_birth_rate = p.birth_rate
 
-        if isinstance(this_birth_rate, ss.TimePar):
-            factor = 1.0
-        else:
-            factor = ss.time_ratio(unit1=self.t.unit, dt1=self.t.dt, unit2='year', dt2=1.0)
-
-        scaled_birth_prob = this_birth_rate * p.rate_units * p.rel_birth * factor
+        scaled_birth_prob = this_birth_rate * self.t.dt * p.rate_units * p.rel_birth
         scaled_birth_prob = np.clip(scaled_birth_prob, a_min=0, a_max=1)
```
daniel-klein, line 92: Multiplication by dt should happen at the end.

daniel-klein, line 92: No need for clipping if `this_birth_rate` is a RateProb (but maybe it's not in some cases?)

starsim/time.py
```diff
@@ -227,70 +108,875 @@ def from_year(cls, year):
...
+
+    def __abs__(self):
+        # Cannot implement this because it's ambiguous how to resolve cases like
+        # DateDur(months=-1,days=1) - this is a sensible DateDur interpreted as 'go back 1 month, then go forward 1 day'
+        # but just taking the absolute value of all of the components wouldn't work because this would be on average 1 month + 1 day
+        # whereas it should be 1 month - 1 day. This could probably be resolved? But is an edge case, unlikely to be needed
+        # (whereas abs(YearDur) arises when comparing dates, which automatically return YearDur)
+        raise NotImplementedError
+
+class Rate():
```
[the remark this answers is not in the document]
robynstuart: Update: after reading ahead and mulling this over for awhile, maybe `Rate` is ok, but `period` should be `unit` I think.

starsim/time.py
```diff
@@ -227,70 +108,875 @@ def from_year(cls, year):
...
+        else:
+            self.period = Dur(period)
+
+        assert sc.isnumber(value), 'Value must be a scalar number'
+        self.value = value
+
+    def __repr__(self):
+        return f'<{self.__class__.__name__}: {self.value} per {self.period.str()}>' # Use str to get the friendly representation
+
+    def __mul__(self, other):
```
robynstuart, line 789: I keep coming back to the example of a waning rate, since there were comments about this above. Let's say you've defined:
```
w = ss.peryear(r)
# Or equivalently:
w = ss.Rate(r, period=ss.years(1))
```
Then I want to multiply this by a duration, e.g. `ss.months(1)` to calculate how much waning has happened over a month. I can see it's valid to multiply a `Rate` by a `Dur`, but this should give `np.exp(-r * factor)` where `factor` is 1/12.

starsim/time.py
```diff
@@ -227,70 +108,875 @@ def from_year(cls, year):
...
+
+    Alternatively, when a ``RateProb`` is multiplied by a duration (type
+    ss.Dur), a probability is calculated. The conversion from rate to
+    probability on multiplication by a duration is
+        ``1 - np.exp(-rate/factor)``,
+    where ``factor`` is the ratio of the multiplied duration to the original
+    period (denominator).
+
+    For example, consider
+    >>> p = ss.RateProb(0.8, ss.years(1))
```
robynstuart, line 926: Could you give a more concrete example of what kind of parameter this would be? I'm still trying to mentally distinguish it from a `Rate`.

tests/test_time.py
```diff
@@ -182,54 +184,152 @@ def test_mixed_timesteps():
...
+    p*Dur(years=2)
+    p * Dur(0.5)
+    p * Dur(months=1)
+
+    p = TimeProb(0.1, Dur(1))
+    p*Dur(years=2)
+    p * Dur(0.5)
+    p * Dur(months=1)
+
+    p = RateProb(0.1, Dur(years=1))
```
robynstuart, line 324: Can we add formal tests to check the behavior of rate*dur, timeProb*dur, etc?

robynstuart (changes requested):
AMAZING amount of work here, so cool to see! I have some questions about the interpretation of `Rate` and `RateProb` in particular, sorry these are coming so late in the game but keen to understand this and make sure it's all operating as we'd expect. Thank you @RomeshA!!!

starsim/time.py
```diff
@@ -227,70 +108,875 @@ def from_year(cls, year):
...
+        """
+        Convert an array of float years into an array of Date instances
+
+        :param array: An array of float years
+        :return: An array of Date instances
+        """
+        return np.vectorize(cls)(array)
+
+    @classmethod
+    def arange(cls, low, high, step=1):
```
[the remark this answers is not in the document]
RomeshA: The step isn't forced to be a duration because the user can specify it as a fraction of a year, but using durations is of course supported - would be using `ss.months(1)` e.g.

```python
>>> ss.Date.arange(2020,2025,ss.months(1))
Out[6]: 
array([<2020.01.01>, <2020.02.01>, <2020.03.01>, <2020.04.01>,
       <2020.05.01>, <2020.06.01>, <2020.07.01>, <2020.08.01>,
       <2020.09.01>, <2020.10.01>, <2020.11.01>, <2020.12.01>,
       <2021.01.01>, <2021.02.01>, <2021.03.01>, <2021.04.01>,
       <2021.05.01>, <2021.06.01>, <2021.07.01>, <2021.08.01>,
       <2021.09.01>, <2021.10.01>, <2021.11.01>, <2021.12.01>,
       <2022.01.01>, <2022.02.01>, <2022.03.01>, <2022.04.01>,
       <2022.05.01>, <2022.06.01>, <2022.07.01>, <2022.08.01>,
       <2022.09.01>, <2022.10.01>, <2022.11.01>, <2022.12.01>,
       <2023.01.01>, <2023.02.01>, <2023.03.01>, <2023.04.01>,
       <2023.05.01>, <2023.06.01>, <2023.07.01>, <2023.08.01>,
       <2023.09.01>, <2023.10.01>, <2023.11.01>, <2023.12.01>,
       <2024.01.01>, <2024.02.01>, <2024.03.01>, <2024.04.01>,
       <2024.05.01>, <2024.06.01>, <2024.07.01>, <2024.08.01>,
       <2024.09.01>, <2024.10.01>, <2024.11.01>, <2024.12.01>,
       <2025.01.01>], dtype=object)
>>> ss.Date.arange(2020,2021,ss.weeks(1))
Out[7]: 
array([<2020.01.01>, <2020.01.08>, <2020.01.15>, <2020.01.22>,
       <2020.01.29>, <2020.02.05>, <2020.02.12>, <2020.02.19>,
       <2020.02.26>, <2020.03.04>, <2020.03.11>, <2020.03.18>,
       <2020.03.25>, <2020.04.01>, <2020.04.08>, <2020.04.15>,
       <2020.04.22>, <2020.04.29>, <2020.05.06>, <2020.05.13>,
       <2020.05.20>, <2020.05.27>, <2020.06.03>, <2020.06.10>,
       <2020.06.17>, <2020.06.24>, <2020.07.01>, <2020.07.08>,
       <2020.07.15>, <2020.07.22>, <2020.07.29>, <2020.08.05>,
       <2020.08.12>, <2020.08.19>, <2020.08.26>, <2020.09.02>,
       <2020.09.09>, <2020.09.16>, <2020.09.23>, <2020.09.30>,
       <2020.10.07>, <2020.10.14>, <2020.10.21>, <2020.10.28>,
       <2020.11.04>, <2020.11.11>, <2020.11.18>, <2020.11.25>,
       <2020.12.02>, <2020.12.09>, <2020.12.16>, <2020.12.23>,
       <2020.12.30>], dtype=object)
``` 

and then the numerical example is

```python
>>>ss.Date.arange(2020,2025,1)
Out[5]: 
array([<2020.01.01>, <2021.01.01>, <2022.01.01>, <2023.01.01>,
       <2024.01.01>], dtype=object)
```

I would be anticipating that the most common usage would just be that default of `1` where the user gets a range of exact years

starsim/networks.py
```diff
@@ -558,8 +558,10 @@ def add_pairs(self):
 
         if isinstance(self.pars.dur, ss.Dist):
             dur = self.pars.dur.rvs(p1)
+        elif self.pars.dur == 0:
+            dur = np.zeros(len(p1))
         else:
-            dur = np.ones(len(p1))*self.pars.dur # Other option would be np.full(len(p1), self.pars.dur.x), but this is harder to read
+            dur = np.ones(len(p1))*self.pars.dur/self.t.dt # Other option would be np.full(len(p1), self.pars.dur.x), but this is harder to read
```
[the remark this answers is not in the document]
RomeshA: The network duration of edges is implemented by the `dur` attribute (of the edge) storing the number of timesteps remaining which is assigned when the edge is created, then in `DynamicNetwork.end_pairs()` at each timestep it subtracts `1` from the edge duration and then removes any edges where the duration value has become <=0. So since that duration is a number of timesteps, `pars.dur` needs to be divided by `self.t.dt` to perform that conversion

starsim/networks.py
```diff
@@ -620,7 +622,7 @@ def add_pairs(self):
         if isinstance(self.pars.dur, ss.Dist):
             dur = self.pars.dur.rvs(p1)
         else:
-            dur = np.ones(len(p1))*self.pars.dur
+            dur = np.ones(len(p1))*self.pars.dur/self.t.dt
```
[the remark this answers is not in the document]
RomeshA: As above, yes `self.pars.dur` is a duration, but `self.edges.dur` is a number of timesteps, hence the conversion

starsim/diseases/sir.py
RomeshA: @daniel-klein I've opened https://github.com/starsimhub/starsim/issues/928 to track this separately

tests/test_time.py
```diff
@@ -30,63 +31,60 @@ def test_classes():
...
+    assert (d5 + d6)/ss.Dur(days=1) == 365.25*2+3
 
     # Test rate units
     rval = 0.7
-    r5 = ss.rate(rval, unit='week').init(parent_unit='day')
-    r6 = ss.rate(rval).init(parent_dt=0.1)
-    assert np.isclose(r5.values, rval/7) # A limitation of this approach, not exact!
-    assert np.isclose(r6.values, rval/10)
+    r5 = ss.Rate(rval, ss.Dur(weeks=1))
+    assert np.isclose(r5*ss.days(1), rval/7) # A limitation of this approach, not exact!
```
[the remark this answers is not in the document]
RomeshA: Looks like this has been resolved by https://github.com/starsimhub/starsim/pull/847/commits/d74285e66d180850562b08318689af51cea08742 but let me know if not

starsim/diseases/hiv.py
```diff
@@ -11,21 +11,21 @@
 
 class HIV(ss.Infection):
 
-    def __init__(self, pars=None, *args, **kwargs):
+    def __init__(self, *args, **kwargs):
         super().__init__()
         self.define_pars(
             unit = 'year',
-            beta = ss.beta(1.0), # Placeholder value
+            beta = ss.TimeProb(1.0), # Placeholder value
```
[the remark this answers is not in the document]
RomeshA: The module isn't supposed to have a unit any more, had missed removing it here :) resolved by removing `unit = 'year'`

starsim/demographics.py
RomeshA: Opened https://github.com/starsimhub/starsim/issues/929 to track separately

starsim/demographics.py
RomeshA: Noted in https://github.com/starsimhub/starsim/issues/929 as this also relates to the interpretation of `rel_birth`

starsim/networks.py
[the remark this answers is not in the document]
robynstuart: I thought we had a (rather informal) convention of using `dur` for parameters/attributes that represent actual calendar time, and then using `ti_*` for things that represent a timestep. I guess we don't have this for networks, but might be nice to have some way of distinguishing these concepts

starsim/demographics.py
RomeshA: Yes I think it would make sense to keep this here because regardless of whether we default to a `Rate` or `RateProb` the user can override it with their own input

docs/tutorials/tut_buildsim.ipynb
```diff
@@ -182,7 +182,7 @@
    },
    "outputs": [],
    "source": [
-    "sim = ss.Sim(start='2020-01-01', stop='2021-01-01', unit='month', diseases='sis', networks=network)  # Simulate from 2020 for 1 year with a monthly timestep\n",
+    "sim = ss.Sim(start='2020-01-01', stop='2021-01-01', dt=ss.Dur(months=1), diseases='sis', networks=network)  # Simulate from 2020 for 1 year with a monthly timestep\n",
```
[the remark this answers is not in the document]
RomeshA: It would be `ss.months(1)` but an open question would be whether it would be useful to also have `ss.month()`/`ss.week()`/`ss.day()` as shortcuts to `ss.months(1)`, `ss.weeks(1)`, `ss.days(1)` - or if having yet another way to do it would be confusing?

docs/tutorials/tut_calibration.ipynb
```diff
@@ -70,17 +70,16 @@
    "source": [
     "def make_sim():\n",
     "    sir = ss.SIR(\n",
-    "        beta = ss.beta(0.075),\n",
+    "        beta = ss.TimeProb(0.075),\n",
```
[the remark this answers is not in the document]
RomeshA: I think it would make sense that there are other processes that could be defined as a `TimeProb` and also that users might want `beta` to be defined differently. I found it quite confusing to have both `beta` being a parameter and `beta` being a type, it was a bit unclear how they related to each other and what combinations are and aren't allowed. So I'd fairly strongly support @daniel-klein's view that we should remove `beta` as a parameter per https://github.com/starsimhub/starsim/issues/846 - a side note is that I think we also need to think a bit more about how the 'transmission probability per contact' and the 'contact rate' get parametrized across the disease and network objects, but that's a separate issue!

starsim/demographics.py
```diff
@@ -60,6 +59,9 @@ def standardize_birth_data(self):
         birth_rate = ss.standardize_data(data=self.pars.birth_rate, metadata=self.metadata)
         if isinstance(birth_rate, (pd.Series, pd.DataFrame)):
             return birth_rate.xs(0,level='age')
+        if sc.isnumber(birth_rate):
+            # If the user has provided a bare number, assume it is per year
```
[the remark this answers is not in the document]
RomeshA: Added in 69b1d20f93f44c847eeda6387df41659e9d79a04

starsim/demographics.py
```diff
@@ -508,7 +507,7 @@ def make_embryos(self, conceive_uids):
             # Add connections to any prenatal transmission layers
             for lkey, layer in self.sim.networks.items():
                 if layer.prenatal:
-                    durs = np.full(n_unborn, fill_value=self.pars.dur_pregnancy)
+                    durs = np.full(n_unborn, fill_value=self.pars.dur_pregnancy/self.t.dt)
```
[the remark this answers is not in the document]
RomeshA: Added a comment

starsim/demographics.py
```diff
@@ -532,15 +531,15 @@ def set_prognoses(self, uids):
         self.ti_pregnant[uids] = ti
 
         # Outcomes for pregnancies
-        dur_preg = np.ones(len(uids))*self.pars.dur_pregnancy
+        dur_preg = self.pars.dur_pregnancy
         dur_postpartum = self.pars.dur_postpartum.rvs(uids)
         dead = self.pars.p_maternal_death.rvs(uids)
-        self.ti_delivery[uids] = ti + dur_preg # Currently assumes maternal deaths still result in a live baby
+        self.ti_delivery[uids] = ti + dur_preg/self.t.dt # Currently assumes maternal deaths still result in a live baby
```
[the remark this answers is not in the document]
RomeshA: Added a comment to this one too. In general, it is just that any time operations are done on time indices (adding to `ti`, storing something in `ti_*`) the durations entered by the user need to get converted to a number of timesteps before performing the operation

starsim/distributions.py
```diff
@@ -769,6 +786,13 @@ def __init__(self, mean=0.0, sigma=1.0, **kwargs): # Does not accept dtype
         super().__init__(distname='lognormal', dist=sps.lognorm, mean=mean, sigma=sigma, **kwargs)
         return
 
+    def convert_timepars(self):
+        for key, v in self._pars.items():
+            if isinstance(v, ss.Dur) or isinstance(v, np.ndarray) and v.shape and isinstance(v[0], ss.Dur):
+                raise NotImplementedError('lognormal_im parameters must be nondimensional')
```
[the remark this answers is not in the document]
RomeshA: They can be array-like, but they can't be `Dur` or `Rate` instances - however, since an array of durations/rates has dtype `object` we can't check if `v.dtype` is `Duration`, instead we have to access an element and check the type of the element. So that's why there is a check for `isinstance(v, np.ndarray)` - but it will only raise an error if it's an array _and_ the element being tested has the wrong type

starsim/loop.py
```diff
@@ -158,10 +158,10 @@ def make_plan(self):
         self.plan = sc.dataframe(raw)
 
         # Sort it by step_order, a combination of time and function order
-        self.plan['step_order'] = self.plan.time + ss.options.time_eps*self.plan.func_order
+        # self.plan['step_order'] = self.plan.time + ss.options.time_eps*self.plan.func_order
```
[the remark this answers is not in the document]
RomeshA: Removed the line entirely, the issue was that adding the `time_eps*func_order` to `time` reduced flexibility with respect to the units, but it's not necessary anyway because now the sorting just uses `self.plan.sort_values(['time','func_order'])` so it just directly sorts first by time, then by `func_order` which I think was the intention all along

starsim/diseases/cholera.py
```diff
@@ -14,31 +14,31 @@ class Cholera(ss.Infection):
...
     Cholera
     """
-    def __init__(self, pars=None, *args, **kwargs):
+    def __init__(self, *args, **kwargs):
         """ Initialize with parameters """
         super().__init__()
         self.define_pars(
             # Initial conditions and beta
-            beta = ss.beta(1.0), # Placeholder value
+            beta = ss.TimeProb(1.0, ss.days(1)), # Placeholder value
```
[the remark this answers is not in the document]
RomeshA: Noted in https://github.com/starsimhub/starsim/issues/846 as part of broader `ss.beta` question

starsim/time.py
```diff
@@ -227,70 +108,875 @@ def from_year(cls, year):
...
+    @property
+    def is_variable(self):
+        """
+        Returns True if the duration has variable length
+
+        Some date-based durations like years and months correspond to a variable
+        period of time. This attribute captures whether or not the quantity has
+        variable length.
+        """
+        raise NotImplementedError
```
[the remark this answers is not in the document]
RomeshA: This one is implemented :) it's just handled by the child classes, so `YearDur` always returns `False` while `DateDur` returns `True` if the duration involves years or months, but not if it only involves weeks and days

starsim/time.py
```diff
@@ -227,70 +108,875 @@ def from_year(cls, year):
...
+        tvec = []
+        t = low
+        while t <= high:
+            tvec.append(t)
+            t += step
+        return np.array(tvec)
+
+
+class YearDur(Dur):
+    # Year based duration e.g., if requiring 52 weeks per year
```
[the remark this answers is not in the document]
RomeshA: It would be one of:

```python
>>> ss.YearDur(4/52)/ss.months(1)
Out[4]: 0.9230769230769231
>>> ss.YearDur(4/52)/ss.years(1/12)
Out[6]: 0.9230769230769231
```

which is I think about what you'd expect if you said the average month had `52/12=4.3333` weeks, so it's essentially `4/4.3333`

starsim/time.py
```diff
@@ -227,70 +108,875 @@ def from_year(cls, year):
...
+                    return self.value # Avoid expensive calculation and precision issues
+                rate = -np.log(1 - self.value)
+                return 1 - np.exp(-rate/factor)
+        else:
+            return self.__class__(self.value*other, self.period)
+
+    def __truediv__(self, other): raise NotImplementedError()
+    def __rtruediv__(self, other): raise NotImplementedError()
+
+class RateProb(Rate):
```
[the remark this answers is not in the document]
RomeshA: I think I understand the issue - an instantaneous rate still needs to be specified with a time dimension in the units e.g., 'kilometres _per hour_' or 'metres _per second_'. A couple of reasons I'd opted for 'period' over 'units' in the are that 

1. the 'unit' of the quantity is 'metres per second', whereas this attribute is only capturing the 'per second' part, so it's not actually the unit of the quantity - and in fact it's the inverse of the time dimension i.e., for 'metres per second' the attribute would be storing 'seconds', not 'per second', so again it's not actually the unit
2. 'Unit' suggests it would be storing just the dimension e.g., 'seconds' as opposed to a specific amount of time '1 second' and the model is expecting the latter - partly because we don't have abstract classes that capture the former, but also because it can make sense to specify a rate per fortnight, or per other period of time, and that gets specified together with whether its seconds, weeks, months etc. 

It's true that the quantity is still instantaneous though, so it's not that it is intrinsically over a period and I can see how that could cause confusion. But I think 'unit' by itself isn't quite right either. Something like 'denominator' might be less ambiguous and more accurate? The other alternative is whether it's something that can be sufficiently clearly explained in the docs

starsim/time.py
RomeshA: Maybe we should have a specific `WaningRate` type?

RomeshA: Ah actually - this is pretty much what `RateProb` does! 

```python
>>> r = 0.5
>>> w = ss.RateProb(r, period=ss.years(1))
>>> w*ss.months(1)
0.040810542890861834
>>> 1-np.exp(-r*1/12)
0.040810542890861834
```

starsim/time.py
RomeshA: As per above example for the waning rate calculation

tests/test_time.py
RomeshA: Done!

starsim/networks.py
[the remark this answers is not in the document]
RomeshA: I like that convention and would support `edges.ti_dur` or something like that for clarity. Alternatively, we could consider storing `self.edges.ti_end` and then remove edges if `self.t.ti >= self.edges.ti_end` rather than decrementing the duration at each timestep, that would probably be more in line with how we handle durations in disease modules?

starsim/time.py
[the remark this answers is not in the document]
RomeshA: (Resolving this comment but see one below)

RomeshA:
@daniel-klein @robynstuart Thanks both for your reviews! I've responded to most of the comments but left a few open for further discussion, although some of those could perhaps be made issues to iterate further independently of this PR

docs/tutorials/tut_demographics.ipynb
```diff
@@ -70,8 +70,8 @@
    "outputs": [],
    "source": [
     "demographics = [\n",
-    "    ss.Births(birth_rate=20),\n",
-    "    ss.Deaths(death_rate=15)\n",
+    "    ss.Births(birth_rate=ss.peryear(20)),\n",
```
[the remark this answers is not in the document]
RomeshA: I think this just works as long as the CSV is in standard (annual) units, which they currently are (e.g., the `time-experiments` branche in `hiv_vmb_kenya`) - we could extend the CSV to support other units, but I think it would make sense for that to be part of a broader refactoring of how we represent input data, it would be nice to also have a way to change whether it's specified per 1000 people or as an absolute number etc. without having separate attributes like `rate_units`

tests/benchmark.json
```diff
@@ -1,12 +1,12 @@
 {
   "time": {
-    "initialize": 0.039,
-    "run": 0.757
+    "initialize": 0.03,
+    "run": 1.417
```
cliffckerr, line 4: Unfortunately I think this performance regression is real ... running `tests/benchmark_full.py`, I do get that things are 2x slower on this branch than `rc3.0`. Specifically, on `rc3.0`, I get:
```
Trial 1/10: 20.0 s
Trial 2/10: 18.9 s
```
and on this branch:
```
Trial 1/10: 48.6 s
Trial 2/10: 49.3 s
```

This is obviously something we'll need to address. I would love to _not_ have to do result caching, but if it's a 2.5x speed improvement, we might need to live with it ...

RomeshA: It might not be too bad depending on how it's implemented, if there's a good place to put it. I'll do some profiling today and investigate a bit more

RomeshA: With the changes just pushed I'm getting about the same runtime on both branches now, let me know if there is still an issue with this!

PR #909, opened 2025-03-28
Add disease=self argument to net.net_beta() function call in infect()

epbennetts:
### Description
-This argument is necessary if using a network class where the Network's net_beta() method has been overridden and which needs access to disease-specific parameters, e.g. if the disease-specific condom efficacy is needed in order to calculate net_beta.
-disease (default=None) is already an argument in the definition of Network.net_beta(), it is just this specific usage that does not include it. 

### Checklist
- [X] Very minor change, no docstrings or version changes needed


RomeshA (approved):
Looks good to me, noting that the parent `Network.net_beta()` already has a `disease` argument, so this is probably just fixing an unintentional omission when calling `net_beta`

Issue #923, opened 2025-04-22
Add `people.filter()`


cliffckerr:
See `fpsim/base.py`. About the only things that should require porting:
- `__getattribute__()`
- `filter()`
- `unfilter()`
- `is_filtered`

cliffckerr:
Done

PR #934, opened 2025-05-12
Rc3.0 filter

cliffckerr:
### Description
Adds FPsim-style filtering to Starsim. **WARNING**, not working yet!

Example test script:
```
import starsim as ss

sim = ss.demo(plot=False)
fpop = sim.people.filter(sim.people.female)
fage = fpop.filter(fpop.age>20)
```

The problem is that the states are global, so while `fage` is filtered correctly, `sim.people` gets filtered too.


cliffckerr:
One possible syntax that should be workable:

```py
import starsim as ss

sim = ss.demo(plot=False)
fpop = sim.people.filter('female')
fage = fpop('age') > 20
fage.inds # The indices to be used
```

robynstuart:
Looks cool! I second @RHull-IDM's that it's not _quite_ as nice as the FPsim example where you get the full people functionality. And the syntax of `f = (sim.people.filter('female') ('age')>5) ('~sir.infected')` is a little involved. Why doesn't this work?
```
ff = sim.people.filter(sim.people.female & (sim.people.age>5) & ~sim.people.sir.infected)
f.age.min()  # Returns 1.013
```

RHull-IDM:
@robynstuart I just tested your snippet and it does seem to work for me, returning 5.00108. I think the syntax of that example is more a stress test of different ways you can group, chain, and specify conditions rather than necessarily the best way.

robynstuart:
Oh you're right, i should've tested `ff.age.min()`! I thought I was missing something.

cliffckerr:
Closes #923

PR #939, opened 2025-05-19
Rc3.0 docs

cliffckerr:
### Checklist


JSchripsema-IDM (approved):
It's unclear to me if this config supports Google Analytics v4, based on the documentation here: https://quarto.org/docs/websites/website-tools.html#google-analytics I think we'll have to push this live to check if the analytics dashboard is capturing data correctly. If not, we may need to insert the container via HTML similar to how we did this with Sphinx.

PR #943, opened 2025-05-29
Cuda support

cliffckerr:
### Description
Adds CUDA (CuPy) support. !. Test scripts:

```py
import starsim as ss

# Static test -- very fast
s1 = ss.Sim(n_agents=500e3, diseases='sis', networks='static', start=0, stop=100)
s1.run()
s1.plot()

# Default test -- very slow since removed Numba support
s2 = ss.Sim(n_agents=10e3, diseases='sis', networks='random', start=0, stop=100)
s2.run()
s2.plot()

### DIFF
--- a/starsim/arrays.py
+++ b/starsim/arrays.py
@@ -353,7 +353,7 @@ class Arr(BaseArr):
         new.__dict__ = self.__dict__.copy() # Copy pointers
         new.dtype = arr.dtype # Set to correct dtype
         new.name = name # In most cases, the asnew Arr has different values to the original Arr so the original name no longer makes sense
-        new.raw = np.empty(new.raw.shape, dtype=new.dtype) # Copy values, breaking reference
+        new.raw = cp.empty(new.raw.shape, dtype=new.dtype) # Copy values, breaking reference
         new.raw[new.auids] = arr
         return new
 
--- a/starsim/arrays.py
+++ b/starsim/arrays.py
@@ -317,7 +317,7 @@ class Arr(BaseArr):
         if orig_len + n_new > self.len_tot:
             n_grow = max(n_new, self.len_tot//2)  # Minimum 50% growth, since growing arrays is slow
             new_empty = cp.empty(n_grow, dtype=self.dtype) # 10x faster than np.zeros()
-            print('TEST', self.name, type(self.raw), type(new_empty))
+            # print('TEST', self.name, type(self.raw), type(new_empty))
             self.raw = cp.concatenate([self.raw, new_empty], axis=0)
             self.len_tot = len(self.raw)
             if n_grow > n_new: # We added extra space at the end, set to NaN
--- a/starsim/arrays.py
+++ b/starsim/arrays.py
@@ -4,6 +4,7 @@ Define array-handling classes, including agent states
 import numpy as np
 import sciris as sc
 import starsim as ss
+import cupy as cp
 
 # Shorten these for performance
 ss_float = ss.dtypes.float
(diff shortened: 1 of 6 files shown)


cliffckerr:
Closing for now, but keeping the branch for ideas

cliffckerr:
Thoughts from Claude on the feasibility of this:

# GPU acceleration feasibility notes

## Bottom line

A universal `ss.options(target='gpu')` drop-in is **not feasible today**, but a narrow **opt-in cupy "turbo mode"** is. The two things we want — a transparent numpy drop-in *and* cross-platform support (CPU + Mac + NVIDIA) — are in direct tension, and no current array library gives both.

## Why drop-in and cross-platform conflict

The property that makes a library a transparent numpy replacement is exactly the property cupy has and torch/jax lack:

| Property `arrays.py` relies on | numpy | cupy | torch / jax |
|---|---|---|---|
| numpy ufuncs (`np.add`, called explicitly in `Arr._math`) stay on-device | ✅ | ✅ (`__array_ufunc__`) | ❌ silently copies to a host `numpy.ndarray` |
| Array is subclassable (`class uids(np.ndarray)`, 14× `.view(cls)`) | ✅ | ✅ | ❌ `.view(subclass)` raises `TypeError` |
| numpy dtype checks (`np.issubdtype(..., np.floating)`) | ✅ | ✅ | ⚠️ torch has its own dtype system |

This is why the cupy PR was ~80 lines while torch would be "a LOT of changes": cupy rides the existing `Arr`/`uids`/`_math` machinery almost untouched; torch requires rewriting the core array class (dispatch ufuncs through a backend handle, replace ndarray subclassing, port the dtype logic). But cupy is NVIDIA-only — it does not solve Mac or CPU.

## Even a perfect backend wouldn't help the common case

GPU only wins for dense arithmetic on very large populations. Quick probe (laptop RTX 3050, so a datacenter card shifts crossovers down, but the shape holds):

- Dense op, 1k / 10k / 100k agents: GPU **0.02× / 0.08× / 0.81×** (slower). Only wins at **1M (2.5×)** and **5M (8.4×)**.
- Realistic loop shape (100 steps, modest work, one scalar pulled to host per step): CPU 0.7 ms vs **GPU 10.3 ms — 15× slower**.
- Irregular gather `arr[idx]` (our dominant `.uids` access pattern): GPU loses in every case tested.

The loop is ~22 Python method dispatches per step, transmission is irregular edge-list gather/scatter (E≈5N) with variable-length outputs, `RandomNet` rebuilds all edges every step, and there are forced GPU→CPU syncs everywhere (`slots.max()+1` on every RNG draw, 8–12 `count_nonzero`/`sum` result writes per step, `if len(uids)` branches). That's why the 6× needed a *static* network + 500k agents; the dynamic (default) case was slow because syncs and gathers dominate. Note also that the natural parallelism here is usually *across* sims (calibration/scenarios), which `MultiSim` already exploits on CPU cores — GPU-within-a-sim only pays off for the "one enormous sim" niche (10M+ agents).

## CRN is a hard blocker — but hash-CRN is the enabler

Reproducibility is welded to PCG64's `.jumped()` + bit-generator state get/set and to scipy.stats `.ppf`/`.rvs` per step, none of which cupy/torch provide (the PR just switched CRN off). The **hash-CRN prototype** is exactly the stateless, counter-based RNG that ports to GPU cleanly — no jump-ahead, no state serialization, embarrassingly parallel. Worth landing on its own merits; it's also the prerequisite that would make any future GPU path reproducible.

## Recommendation

1. **Don't build the universal `target='gpu'` drop-in.** No library satisfies drop-in + cross-platform, and the workload resists GPU below millions of agents / for dynamic networks. A torch rewrite risks *regressing* the CPU case everyone uses to help a rare one.
2. **Keep investing in CPU** (numba/vectorization wins, `MultiSim` across cores) — higher leverage for how Starsim is actually used.
3. **Land hash-CRN independently.** It's the enabler for a reproducible GPU path later.
4. **Only if a real user needs a huge single static-network sim on NVIDIA:** ship a lazy, opt-in cupy backend via a thin `starsim/backend.py` (`xp` handle flipped by the existing `refresh_references()` hook), clearly scoped as narrow turbo mode. Lazy import means no new hard dependency and nothing breaks on Macs/CPU-only machines.

PR #972, opened 2025-07-03
Update README.rst

MIzzo-IDM:
made some edits to the readme, just minor language tweaks and added some missing commas. Note that I changed "diseases" to "systems" in the second paragraph, since you're including malnutrition (systems may not be the best choice, but wanted something broader than diseases...). Also, should the Analyzers class have its own entry in the submodules list? It's currently listed with interventions.py but it sounds like it should be a separate entry.

### Checklist

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,9 +1,9 @@
 Starsim
 =======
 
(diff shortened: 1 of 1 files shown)


MIzzo-IDM:
updated the wrong file (did not update the current working branch)! Closing this as the correct readme has been updated.

PR #975, opened 2025-07-07
Speed up time.py slightly

kelvinburke:
### Description
A couple speed ups:
They rely on the fact that `pd.Timestamp` is immutable, and we can change from a `pd.Timestamp` to a `ss.date` by calling `ss.date._reset_class(timestamp)`. This means when we create a pd.Timestamp in the functions `ss.date.from_year()` and `ss.date.__add__()` we don't need to copy the `pd.Timestamp` we get (which would happen in `ss.date.__new__()` ) - we only need to convert it from `pd.Timestamp` into a `ss.date` .

The other change is to not re-create the `ss.Time._yearvec` and `ss.Time._tvec` of each module if it is the same `start, stop, dt` as the `sim.t` - we can just create a copy. I guess this could break if the `sim.t` was initialized - then any of the `sim.t.start, sim.t.stop, sim.t.dt, sim.t._yearvec, sim.t._tvec` were then manually changed without reinitializing, then the module `time` objects were initialized, but that is not really an intended use.

### Checklist

### DIFF
--- a/starsim/time.py
+++ b/starsim/time.py
@@ -1291,6 +1291,14 @@ class Time:
             if self.is_absolute != sim.t.is_absolute:
                 raise Exception('Cannot mix absolute/relative times across modules')
 
+        if sim.t.initialized and \
+                all([type(getattr(self, key)) == type(getattr(sim.t, key)) for key in ('start', 'stop', 'dt')]) and \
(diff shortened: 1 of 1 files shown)


cliffckerr (approved):
Looks great, thank you!

PR #976, opened 2025-07-07
Rc3.0 intvs

robynstuart:
### Description
Remove HPV-specific interventions from Starsim

### Checklist


cliffckerr (commented):
Happy to merge, but you're sure these wouldn't be useful to have as examples? Could also move to the `starsim_examples` folder?

robynstuart:
@cliffckerr I'm in the process of moving these all to [HPVsim](https://github.com/starsimhub/hpvsim/tree/rc3-init) so I thought it'd be a good chance to clear them out of here. Are there any other spokes that use them?

Issue #998, opened 2025-07-11
Swap to using  `timeprob`  in  `ss.Deaths.make_death_prob_fn()`.


kelvinburke:
Hi @cliffckerr 

I was looking at `profile_sim.py` with different types of sims and noticed that there can be some speed ups in `ss.Deaths.make_death_prob_fn()`.

But more importantly - it might need to be updated to use either `timeprob` or `rateprob`. Currently the main line in `ss.Deaths.make_death_prob_fn()` is:

```
death_prob = death_rate * self.t.dt * self.pars.rate_units * self.pars.rel_death
```
where the types are generally
```
death_rate = array of float or single float or single Rate (any type)
self.t.dt = Dur either YearDur or DateDur
self.pars.rate_units = constant float
self.pars.rel_death = constant float
```
this means when `death_rate = array of float` , then `death_prob = array of Dur` which isn't really what I'd expect

But really I think it makes sense for `death_rate` to be a `timeprob`? That's how I normally interpret demographics data is a probability of death over 1 year.

So then it would be 
```
death_prob = (death_rate * self.pars.rate_units * self.pars.rel_death).to_prob(self.t.dt)
```
Does that make sense?

PR #999, opened 2025-07-11
Swap demographics to using `timeprob`

kelvinburke:
### Description

Draft
Haven't checked the numbers are the same (they probably won't be exactly the same because we are scaling with the exponential according to `timeprob.to_prob()`)

### Checklist


kelvinburke:
@cliffckerr   Would be the fix for https://github.com/starsimhub/starsim/issues/998, and it is faster too

cliffckerr (commented):
Thanks! Will think about this, am changing the way time is implemented so not sure if this implementation will also need to change, but should have an answer soon :)

PR #1000, opened 2025-07-11
Refactor arrays.py slightly

kelvinburke:
### Description

Hi @cliffckerr 

Sorry this PR is into your working branch because it involves the `nb_indexer`.

Changes are
- Add the check for the `_numba_threshold` to `Arr.grow` so that `People.slot` can also use the `nb_indexer` (previously it wasn't because it doesn't get `link_people` called.
- Remove the if statement in `Arr._index` by splitting into two wrappers `Arr._np_index, _nb_index` (need to wrap because of that sciris issue)

But its a bit faster (mainly because `slot` can use `nb_indexer`) and a bit neater

### Checklist


cliffckerr:
Woo lucky 1000! :)
image

cliffckerr (approved):
Small edits but looks great, thank you!

PR #1031, opened 2025-08-12
Fix indexing

robynstuart:
### Checklist


starsim/demographics.py
```diff
@@ -410,6 +413,7 @@ def make_fertility_prob_fn(self, sim, uids):
 
         # Scale from rate to probability
         invalid_age = (age < self.pars.min_age) | (age > self.pars.max_age)
+        fertility_rate = np.clip(fertility_rate, a_min=0, a_max=1-ss.options.time_eps)  # Fertility rates should not be >1 or <0
```
robynstuart, line 416: I realize this isn't super elegant, but otherwise I get warnings from converting to probs... and realistically, we should not be seeing an annual probability of >1 here

starsim/demographics.py
```diff
@@ -383,7 +383,10 @@ def make_fertility_prob_fn(self, sim, uids):
         if isinstance(frd, ss.Rate):
             fertility_rate[uids] = ss.prob(frd.value * (self.pars.rate_units * self.pars.rel_fertility), frd.unit).to_prob(ss.years(1))
         else:
-            year_ind = sc.findnearest(frd.index, self.t.now('year')-self.pars.dur_pregnancy.years)
+            conception_time = self.t.now('year')-self.pars.dur_pregnancy.years
+            if isinstance(conception_time, ss.years):
+                conception_time = conception_time.years
+            year_ind = sc.findnearest(frd.index, conception_time)
```
cliffckerr, line 389: `self.t.now('year')` should return a floating point year always ... that would fix this, right? Related to #1033

cliffckerr: `self.t.year` is now implemented on #1032

starsim/demographics.py
```diff
@@ -198,7 +198,7 @@ def __init__(self, pars=None, rel_death=_, death_rate=_, rate_units=_, metadata=
         # Process data, which may be provided as a number, dict, dataframe, or series
         # If it's a number it's left as-is; otherwise it's converted to a dataframe
         self.death_rate_data = self.standardize_death_data() # TODO: refactor
-        self.pars.death_rate = ss.bernoulli(p=self.make_death_prob_fn)
+        self.pars.p_death = ss.bernoulli(p=0)  # Placeholder, populated by make_p_death
```
robynstuart, line 201: Renaming to better represent how it gets used

starsim/demographics.py
```diff
@@ -215,10 +215,11 @@ def standardize_death_data(self):
             return ss.peryear(death_rate)
         return death_rate
 
-    @staticmethod # Needs to be static since called externally, although it sure looks like a class method!
-    def make_death_prob_fn(self, sim, uids):
+    def make_p_death(self):
```
robynstuart, line 218: I think this should be a class method... have adjusted

Issue #1073, opened 2025-11-06
Pregnancy uplift


robynstuart:
- [ ] assign variable `dur_pregnancy`, using data-based distributions on gestation time
- [ ] use dur_pregnancy to vary delivery time
- [ ] add `gestation` attribute that tracks gestational time for currently-pregnant women
- [ ] add `gestation_at_birth` attribute for gestational time of newborn agents (nan if not conceived during sim)
- [ ] add `fertile` state along with `_p_fertile` (copy from fp)
- [ ] think about hooks for contraception integration... 
- [ ] add rel_ptb 
- [ ] link rel_ptb to maternal age

PR #1074, opened 2025-11-06
Initial pregnancy uplift

robynstuart:
### Description
https://github.com/starsimhub/starsim/issues/1073

### Checklist

### DIFF
--- a/starsim/demographics.py
+++ b/starsim/demographics.py
@@ -393,6 +393,7 @@ class Pregnancy(Demographics):
             ss.FloatArr('dur_pregnancy', label='Pregnancy duration'),  # Duration of pregnancy
             ss.FloatArr('parity', label='Parity', default=0),  # Number of births (may include live + still)
             ss.FloatArr('n_pregnancies', label='Number of pregnancies', default=0),  # Number of pregnancies
+            ss.FloatArr('n_miscarriages', label='Number of miscarrianges', default=0),  # Number of micarriages
             ss.FloatArr('gestation', label='Number of weeks into pregnancy'),  # Gestational clock
             ss.FloatArr('gestation_at_birth', label='Gestational age in weeks'),  # Gestational age at birth, NA if not born during the sim
             ss.FloatArr('ti_pregnant', label='Time of pregnancy'),  # Time pregnancy begins
@@ -452,22 +453,6 @@ class Pregnancy(Demographics):
     def nulliparous(self):
         """ Women of childbearing age who have never been pregnant """
         return (self.parity == 0) & self.fecund
-
-    # NB - consider moving to modules that check postpartum time? Means postpartum period can be
-    # specific to each module
-    #
-    # @property
-    # def postpartum(self):
-    #     """
-    #     Within a pre-defined postpartum window, as specified by the dur_postpartum par
-    #     This does not directly affect any other functionality within this module, but is
-    #     provided for convenience for modules that need to know which women are X timesteps
-    #     postpartum.
-    #     """
-    #     timesteps_since_birth = self.ti - self.ti_delivery
-    #     pp_timesteps = int(self.pars.dur_postpartum/self.t.dt)
-    #     pp_bools = ~self.pregnant & (timesteps_since_birth >= 0) & (timesteps_since_birth <= pp_timesteps)
-    #     return pp_bools
 
     @property
     def dur_gestation(self):
@@ -714,7 +699,7 @@ class Pregnancy(Demographics):
 
         # Set maternal death outcomes
         dead = self.pars.p_maternal_death.filter(mother_uids)
-        self.ti_dead[mother_uids[dead]] = self.ti
+        self.ti_dead[dead] = self.ti
 
         # Update networks after the states have been updated above so that the
         # networks can access the latest/current states
@@ -729,6 +714,13 @@ class Pregnancy(Demographics):
             if isinstance(net, ss.PostnatalNet):
                 net.add_pairs(newborn_parents, newborn_uids)
 
+        # Hook for subclasses to add post-delivery logic
+        self._post_delivery(mother_uids, newborn_uids)
+
+        return
+
+    def _post_delivery(self, mother_uids, newborn_uids):
+        """Hook for subclasses to add post-delivery logic. Base implementation does nothing."""
         return
 
     def set_breastfeeding(self, mother_uids, newborn_uids):
@@ -857,10 +849,7 @@ class Pregnancy(Demographics):
 
     def make_pregnancies(self, uids):
         """
-        Make pregnancies
-        Add miscarriage/termination logic here
-        Also reconciliation with birth rates
-        Q, is this also a good place to check for other conditions and set prognoses for the fetus?
+        Make pregnancies, assign durations, and determine multiples
         """
 
         # Change states for the newly pregnant woman
@@ -942,9 +931,15 @@ class Pregnancy(Demographics):
         death_uids = ss.uids(self.sim.people.ti_dead <= self.ti)
         if len(death_uids) == 0:
             return
+        self.process_maternal_deaths(death_uids)
+        self.process_prenatal_deaths(death_uids)
+        return
 
-        # Any pregnant? Consider death of the unborn agent. Default probability is set to 1
-        # meaning we assume that unborn children do not survive.
+    def process_maternal_deaths(self, death_uids):
+        """
+        If any pregnant mothers die, we also process the death of the unborn agent.
+        Default probability is set to 1 meaning we assume that unborn children do not survive.
+        """
         mother_death_uids = death_uids[self.pregnant[death_uids]]
         if len(mother_death_uids):
             unborn_uids = self.find_unborn_children(mother_death_uids)
@@ -952,17 +947,19 @@ class Pregnancy(Demographics):
             if len(unborn_death_uids):
                 self.sim.people.request_death(unborn_death_uids)
             self.step_die(mother_death_uids)
+        return
 
-        # Any prenatal? Handle changes to pregnancy
+    def process_prenatal_deaths(self, death_uids):
+        """
+        If any deaths occur in unborn agents, we handle changes to pregnancy and record as a miscarriage
+        """
         is_prenatal = self.sim.people.age[death_uids] < 0
         prenatal_death_uids = death_uids[is_prenatal]
         if len(prenatal_death_uids):
             mother_uids = self.sim.people.parent[prenatal_death_uids]
-
-            # Figure out whether the mother is still pregnant, which is possible if she was carrying multiple embryos
+            self.n_miscarriages[mother_uids] += 1
             singletons = mother_uids[~self.carrying_multiple[mother_uids]]
             self.step_die(singletons)
-
         return
 
     def update_results(self):


starsim/demographics.py
```diff
@@ -324,12 +324,16 @@ def __init__(self, pars=None, dur_pregnancy=_, dur_postpartum=_, fertility_rate=
...
-            fertility_rate = ss.peryear(100), # Can be a number of Pandas DataFrame
+            dur_pregnancy = ss.choice(a=ss.weeks(np.arange(32, 43)), p = np.array([0.001, 0.002, 0.005, 0.012, 0.026, 0.05, 0.087, 0.134, 0.188, 0.226, 0.269])), # Quantiles for looking up fertility rates at delivery time
+            rel_ptb_under18 = 1.2, # Relative risk of pre-term birth for mothers under 18
+            rel_ptb_over35 = 1.2, # Relative risk of pre-term birth for mothers over 35
+            fertility_rate = ss.peryear(100), # Can be a number or a Pandas DataFrame
             rel_fertility = 1,
+            primary_infertility = 0,  # Probability of primary infertility
             p_maternal_death = ss.bernoulli(0),
-            p_neonatal_death = ss.bernoulli(0),
+            p_neonatal_death = ss.bernoulli(1),
```
jamiecohen: 100%?!

starsim/demographics.py
```diff
@@ -340,35 +344,73 @@ def __init__(self, pars=None, dur_pregnancy=_, dur_postpartum=_, fertility_rate=
...
+        self.asfr_width = self.asfr_bins[1]-self.asfr_bins[0]
+        self.asfr = None  # Storing this separately from results as it has a different format
+
         return
 
-    def make_p_fertility(self, filter_uids=None):
+    @property
+    def trimesters(self):
+        """ Return trimester durations in weeks """
+        return [ss.weeks(13), ss.weeks(26), ss.weeks(40)]
```
jamiecohen: do you ever use the ss.weeks(40) value? might not need

starsim/demographics.py
robynstuart: This is the probability that a baby in utero will survive the death of the mother. We could certainly make this more precise and link it to the gestational age, but I made this change because in most circumstances, this would be close to 1 unless the baby was extremely close to birth and the death of the mother occurred in the context of care.It was 0 before, which meant that we had agents aged -0.5 years old that survived...

starsim/demographics.py
robynstuart: No but we could potentially use it in the context of e.g. ANC interventions

starsim/demographics.py
```diff
@@ -340,35 +344,73 @@ def __init__(self, pars=None, dur_pregnancy=_, dur_postpartum=_, fertility_rate=
...
+        self._p_fertile = ss.bernoulli(p=1-self.pars['primary_infertility'])  # Probability that a woman is fertile
+        self._p_miscarriage = ss.bernoulli(p=0)  # Probability of miscarriage - placeholder, not used
+        self._p_conceive = ss.bernoulli(p=0)   # Placeholder, see make_p_conceive
+        self._p_stillbirth = ss.bernoulli(p=0)  # Probability of stillbirth - placeholder, not used
+        self._p_twins = ss.bernoulli(p=0)  # Probability of twins - placeholder, not used
 
         # Other, e.g. postpartum, on contraception...
         self.define_states(
             ss.BoolState('fecund', default=True, label='Female of childbearing age'),
+            ss.BoolState('fertile', label='Fertile'),  # Fecund and not infertile
```
ctedijanto: where does this get applied?

starsim/demographics.py
```diff
@@ -340,35 +344,73 @@ def __init__(self, pars=None, dur_pregnancy=_, dur_postpartum=_, fertility_rate=
...
+        # Distributions: binary outcomes
+        self._p_fertile = ss.bernoulli(p=1-self.pars['primary_infertility'])  # Probability that a woman is fertile
+        self._p_miscarriage = ss.bernoulli(p=0)  # Probability of miscarriage - placeholder, not used
+        self._p_conceive = ss.bernoulli(p=0)   # Placeholder, see make_p_conceive
+        self._p_stillbirth = ss.bernoulli(p=0)  # Probability of stillbirth - placeholder, not used
+        self._p_twins = ss.bernoulli(p=0)  # Probability of twins - placeholder, not used
 
         # Other, e.g. postpartum, on contraception...
         self.define_states(
             ss.BoolState('fecund', default=True, label='Female of childbearing age'),
```
RomeshA: Noting that I haven't looked extensively at usage in literature, I have the impression that this could be confusing if fecundity is referring solely to reproductive age and the flag would be true even if they were infertile? It might be clearer if the reproductive/childbearing age state name was more explicitly tied to the age. Alternatively it could be fecund that represents both reproductive age and not infertile

Also if it is related to age, would it make sense for it to be a property that's dynamically computed based on `min_age` and `max_age`?

starsim/demographics.py
RomeshA: Depending on the performance - structurally I think it might be preferable to have an infertile flag rather than a fertile flag, so that it could be independent of fecundity/reproductive age rather than incorporating both. In which case there could be states for `age` and `infertile`, pars for `min_age` and `max_age` and then

`@property fecund` - `age>=min_age & age<=max_age` (notwithstanding the other comment about renaming this)
`@property fertile` - `fecund & ~infertile`

starsim/demographics.py
```diff
@@ -340,35 +344,73 @@ def __init__(self, pars=None, dur_pregnancy=_, dur_postpartum=_, fertility_rate=
...
+            ss.BoolState('fertile', label='Fertile'),  # Fecund and not infertile
             ss.BoolState('pregnant', label='Pregnant'),  # Currently pregnant
             ss.BoolState('postpartum', label="Post-partum"),  # Currently post-partum
             ss.FloatArr('parity', label='Parity', default=0),  # Number of pregnancies
             ss.FloatArr('child_uid', label='UID of children, from embryo through postpartum'),
+            ss.FloatArr('dur_pregnancy', label='Pregnany duration'),  # Duration of pregnancy
```
RomeshA: Typo in label

starsim/demographics.py
RomeshA: Any particular reason for these to be defined inside a function rather than having them be pars or class variables?

starsim/demographics.py
```diff
@@ -340,35 +344,73 @@ def __init__(self, pars=None, dur_pregnancy=_, dur_postpartum=_, fertility_rate=
...
             ss.FloatArr('dur_postpartum', label='Post-partum duration'),  # Duration of postpartum phase
+            ss.FloatArr('gestation', label='Number of weeks into pregnancy'),  # Gestational clock
```
RomeshA: Could/should the gestation be a property attribute?

starsim/demographics.py
```diff
@@ -580,7 +664,7 @@ def make_embryos(self, conceive_uids):
             # Add connections to any prenatal transmission layers
             for lkey, layer in self.sim.networks.items():
                 if layer.prenatal:
-                    durs = np.full(n_unborn, fill_value=self.pars.dur_pregnancy/layer.t.dt) # Network edge duration is stored in units of timesteps, so we need to convert dur_pregnancy to a number of timesteps here
+                    durs = self.dur_pregnancy[conceive_uids]
                     start = np.full(n_unborn, fill_value=self.ti)
                     layer.add_pairs(conceive_uids, new_uids, dur=durs, start=start)
```
RomeshA: It is a bit redundant/fragile to have to synchronize the pregnancy duration across both the pregnancy module and the network, given that the pregnancy duration could change due to interventions. It would be nice if prenatal network could be bound to the pregnancy module, and then to have the network itself compute the edges based on the current pregnancies. But not sure if that would be prohibitively expensive to compute, is that something we've tried before?

starsim/demographics.py
RomeshA: It feels a bit unclear to me how the ss.weeks(40) value relates to the pregnancy duration, are there any particular use cases where the pregnancy duration wouldn't be enough?

starsim/distributions.py
```diff
@@ -1077,9 +1080,9 @@ def __init__(self, mean=0.0, sigma=1.0, **kwargs): # Does not accept dtype
     def convert_timepars(self):
         for key, v in self._pars.items():
             if isinstance(v, ss.dur) or isinstance(v, np.ndarray) and v.shape and isinstance(v[0], ss.dur):
-                raise NotImplementedError('lognormal_im parameters must be nondimensional')
+                raise NotImplementedError('lognormal parameters must be nondimensional')
```
RomeshA: Maybe it would be clearer if this was `lognormal_im` as it's not a problem for `lognormal_ex`, and we probably do want to make sure that users know there is a difference between the two?

A side note is that we probably need to clarify this error message that for lognormal, the parameters can't be dimensional in a mathematical sense (because they get exponentiated) not just a code/class sense

starsim/distributions.py
```diff
@@ -1284,9 +1287,9 @@ def __init__(self, *args, low=None, high=None, dtype=ss.dtypes.rand_int, allow_t
     def convert_timepars(self):
         for key, v in self._pars.items():
             if isinstance(v, ss.dur) or isinstance(v, np.ndarray) and v.shape and isinstance(v[0], ss.dur):
-                raise NotImplementedError('lognormal_im parameters must be nondimensional')
+                raise NotImplementedError('randint parameters must be nondimensional')
```
RomeshA: A side note is that we probably could lift this restriction? It would make sense to be able to sample an integer number of timesteps up to some range, it's not really different to rounding a uniform distribution is it

starsim/demographics.py
robynstuart: Have renamed to `p_survive_maternal_death` and returned it to a default value of 0

starsim/demographics.py
```diff
@@ -466,37 +517,70 @@ def init_results(self):
...
+        return
 
-    def do_step(self):
-        """ Perform all updates """
-        self.update_states()
-        conceive_uids = self.make_pregnancies()
-        self.n_pregnancies += len(conceive_uids) # += to handle burn-in
-        new_uids = self.make_embryos(conceive_uids)
-        return new_uids
+    def set_states(self, uids=None, upper_age=None):
```
RomeshA: It's a little unclear what this function is meant to do based on the name and the arguments given that it's never called with any arguments, would it be OK to remove the arguments from this function, maybe combine it with `set_ptb` and thus also remove `_get_uids`? Or is this function being called from somewhere else with the `uids` and/or `upper_age` argument being provided?

starsim/demographics.py
```diff
@@ -602,44 +686,88 @@ def set_prognoses(self, uids):
...
+        self.gestation[uids] = 0
 
-        # Outcomes for pregnancies
-        dur_preg = self.pars.dur_pregnancy
+        # Use rel_ptb to assign pregnancy durations
+        rel_ptb = self.rel_ptb[uids]
+        sorted_uids = uids[np.argsort(-rel_ptb)]
+        dur_preg = self.pars.dur_pregnancy.rvs(uids)
+        self.dur_pregnancy[sorted_uids] = np.sort(dur_preg)
+        self.ti_delivery[sorted_uids] = ti + self.dur_pregnancy[sorted_uids]  # Currently assumes maternal deaths still result in a live baby.
```
RomeshA: This comment is an important point, we should have an obvious place that users can extend a function to calculate birth outcomes. A `Pregnancy.deliver(maternal_uids)` method would make sense for orchestrating actions that need to take place upon delivery, but it might also make sense to further separate it into something that corresponds specifically for birth outcomes. 


A minimal example of what could go into that function could be setting neonatal survival based on gestational age at the time of delivery/maternal death

RomeshA (commented):
Thanks @robynstuart this is looking great! I've left a few comments on the structure and some functions we could introduce to provide some extension points for child classes. I think there are some redundancies/duplication in the states that could be simplified by dynamically calculating some of the quantities rather than updating them, but whether that makes sense or not would depend on downstream usage as there would be a performance penalty if they prove expensive and frequently accessed, so hopefully that's something that we can investigate even if we don't end up actually going down that route

starsim/demographics.py
robynstuart: Updates: I've added `fecund`, `fertile`, and `susceptible` as properties

starsim/demographics.py
robynstuart: I've tested this and it does seem like performance is better if we store it as a state.

starsim/demographics.py
```diff
@@ -340,35 +344,73 @@ def __init__(self, pars=None, dur_pregnancy=_, dur_postpartum=_, fertility_rate=
...
-    def make_p_fertility(self, filter_uids=None):
+    @property
+    def trimesters(self):
+        """ Return trimester durations in weeks """
+        return [ss.weeks(13), ss.weeks(26), ss.weeks(40)]
+
+    @property
+    def tri1(self):
+        """ Return boolean array for first trimester """
+        return self.gestation < self.trimesters[0].value
```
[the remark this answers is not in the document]
robynstuart: So I've added a new property `dur_gestation` which is a dynamically-computed array rather than a floatarr, and set it up so the `gestation` floatarr is just the dynamic version converted to weeks. I think it will be useful to keep the floatarr version

starsim/demographics.py
robynstuart: Added comments and fixed a mistake in how it was being called on each timestep

starsim/demographics.py
robynstuart: I'll make an issue for this!

starsim/demographics.py
robynstuart: moved

starsim/modules.py
```diff
@@ -160,6 +160,40 @@ def wrapper(self, *args, **kwargs):
...
+
+    This decorator creates a property that:
+    1. Calculates/returns values dynamically like a normal property
+    2. If the returned value is a state that isn't registered, registers it automatically
+
+    Notes:
+    - Acts like a normal @property decorator
+    - If the returned value is an unregistered state, it gets added to the module
+    - Use for computed states or lazy state initialization
+    - Questions: will this be slow?? Why not just register the state at the beginning? What's the advantage
```
RomeshA: I guess this would mainly be useful when people want to make a result that counts the computed property? I think it's helpful for it not to be *too* much of a real state, in the sense that there are things like growing the state arrays that are only relevant for actual states, and we don't want to make it confusing whether something is a real state or a property state. 

All of which to say - I think it could be helpful for users when wanting to produce results from a computed state, but I'm not sure it should be 'registered' alongside other real states.

If this isn't actually used in the current PR, should we separate this out into its own draft PR for later?

starsim/demographics.py
```diff
@@ -5,12 +5,15 @@
 import starsim as ss
 import sciris as sc
 import pandas as pd
+import numba as nb
```
RomeshA: I think we don't need this import any more since you've got find_children working with np.isin?

starsim/demographics.py
```diff
@@ -306,123 +347,221 @@ class Pregnancy(Demographics):
...
+        return self.pregnant.uids[self.dur_gestation >= self.trimesters[1]]
+
+    @staticmethod
+    def _find_children(child_uids, parent_uids, target_parent_uids):
+        """Find children whose parents are in the target set"""
+        mask = np.isin(parent_uids, target_parent_uids)
+        return child_uids[mask]
+
+    def find_children(self, parent_uids):
+        children = self._find_children(self.sim.people.auids, self.sim.people.parent, parent_uids)
```
RomeshA: I think you could put the contents of `_find_children` here instead of having a separate `_find_children` function - that was only required in my original suggestion because `numba` won't let you return an `ss.uids` instance so it needs a standalone function that returns a numpy type

starsim/demographics.py
```diff
@@ -599,20 +874,82 @@ def set_prognoses(self, uids):
...
-        self.ti_delivery[uids] = ti + dur_preg/self.t.dt # Currently assumes maternal deaths still result in a live baby. Note that since we are adding the duration to `ti` (the number of timesteps) we need to convert `dur_preg` into a number of timesteps
-        self.ti_postpartum[uids] = self.ti_delivery[uids] + dur_postpartum
-        self.dur_postpartum[uids] = dur_postpartum
+        return
 
-        if np.any(dead): # NB: 100x faster than np.sum(), 10x faster than np.count_nonzero()
-            self.ti_dead[uids[dead]] = ti + dur_preg
+    def step(self):
+        if self.ti == 0 and self.pars.burnin:  # TODO: refactor
+            dtis = np.arange(np.ceil(-1 * self.pars.dur_pregnancy.rvs()), 0, 1).astype(int)
```
RomeshA: I wonder if it would make sense to do this as part of `init_post`?

starsim/demographics.py
```diff
@@ -306,123 +347,221 @@ class Pregnancy(Demographics):
...
+            ss.FloatArr('dur_pregnancy', label='Pregnancy duration'),  # Duration of pregnancy
+            ss.FloatArr('parity', label='Parity', default=0),  # Number of births (may include live + still)
+            ss.FloatArr('n_pregnancies', label='Number of pregnancies', default=0),  # Number of pregnancies
+            ss.FloatArr('gestation', label='Number of weeks into pregnancy'),  # Gestational clock
+            ss.FloatArr('gestation_at_birth', label='Gestational age in weeks'),  # Gestational age at birth, NA if not born during the sim
             ss.FloatArr('ti_pregnant', label='Time of pregnancy'),  # Time pregnancy begins
-            ss.FloatArr('ti_delivery', label='Time of delivery'),  # Time of delivery
-            ss.FloatArr('ti_postpartum', label='Time post-partum ends'),  # Time postpartum ends
+            ss.FloatArr('ti_delivery', label='Timestep of delivery'),  # Timestep of delivery (integer)
+            ss.FloatArr('date_delivery', label='Date of delivery'),  # Date of delivery (ss.DateArray)
```
RomeshA: I wonder if we need this? Since users should be able to use the `ti_delivery` to index the timeline to get the date, if they need it?

starsim/demographics.py
```diff
@@ -447,103 +586,237 @@ def init_pre(self, sim):
...
-        maternal_deaths = (self.ti_dead <= ti).uids
+    def progress_pregnancies(self):
+        """
+        Update pregnant women. The method can be enhanced by derived classes that add logic
+        for miscarriage, termination, maternal death, etc.
+        """
+        # Update gestational clock for ongoing pregnancies that aren't going to deliver on this timestep
+        if self.pregnant.any():
+            will_deliver = (self.ti_delivery > self.ti) & (self.ti_delivery < (self.ti+1)) & self.pregnant
+            self.gestation[self.pregnant] = self.dur_gestation.weeks
```
RomeshA: Since this is an actual state, probably more efficient to just increment it by dt in weeks rather than using `dur_gestation`?

starsim/demographics.py
```diff
@@ -447,103 +586,237 @@ def init_pre(self, sim):
...
+            self.gestation[will_deliver] = (self.t.dt*self.dur_pregnancy[will_deliver]).weeks  # Set to full term for those delivering this timestep
```
RomeshA: This comment is a bit confusing I think - the idea here is that it's being set to the gestational age at delivery which might be part-way through the timestep (so not that it makes the pregnancy a full-term pregnancy or anything like that)

starsim/demographics.py
```diff
@@ -447,103 +586,237 @@ def init_pre(self, sim):
...
+
+        # Check that gestational clock has values for all currently-pregnant women
+        if np.any(self.pregnant & np.isnan(self.gestation)):
```
RomeshA: If this hasn't caused problems - my feeling is that we probably don't need this check?

starsim/demographics.py
```diff
@@ -306,123 +347,221 @@ class Pregnancy(Demographics):
...
+    def find_children(self, parent_uids):
+        children = self._find_children(self.sim.people.auids, self.sim.people.parent, parent_uids)
+        return ss.uids(children)
+
+    def find_unborn_children(self, parent_uids):
+        all_children = self.find_children(parent_uids)
+        unborn_children = all_children[self.sim.people.age[all_children] < self.t.dt.years]
+        return unborn_children
+
+    def make_p_conceive(self, filter_uids=None):
```
RomeshA: Kind of random comment - this naming scheme is used in a couple of places (like `make_p_death` as well) but I keep getting confused by it, because the probability can be captured by a dist or a function, so the naming often makes me think that `make_p_conceive` is going to return an `ss.Dist` that can be used as `self.p_conceive`, rather than being the probability values. What do you think about naming these (and similar related ones) something like `calc_p_conceive` instead?

starsim/demographics.py
```diff
@@ -447,103 +586,237 @@ def init_pre(self, sim):
...
         # People eligible to become pregnant. We don't remove pregnant people here, these
         # are instead handled in the fertility_dist logic as the rates need to be adjusted
-        eligible_uids = self.sim.people.female.uids
-        p_fertility = self.make_p_fertility(eligible_uids)
-        self.pars.p_fertility.set(p_fertility)
-        dist = self.pars.p_fertility
-        conceive_uids = dist.filter(eligible_uids)
+        if uids is None: uids = self.sim.people.female.uids
+        p_conceive = self.make_p_conceive(uids)
+        self._p_conceive.set(p_conceive)
```
RomeshA: Out of interest, in other places we're sometimes able to just do `self._p_conceive = ss.bernoulli(p=make_p_conceive)` and have the dist automatically call the function to update the probabilities, rather than explicitly calling `Dist.set()` - could that be done here, or is there something different about `make_p_conceive` that prevents it from being done that way?

starsim/demographics.py
```diff
@@ -447,103 +586,237 @@ def init_pre(self, sim):
...
+        return
+
+    def update_prenatal_network(self, conceive_uids, new_uids):
+        """ Add connections to any pre-natal networks """
+
+        # Add connections to any prenatal transmission layers
         for lkey, layer in self.sim.networks.items():
-            if layer.postnatal and self.n_births:
+            if layer.prenatal:
+                durs = self.dur_pregnancy[conceive_uids]
```
RomeshA: Demographics get stepped before networks so maybe it would be OK if the networks (both prenatal and postnatal) just read the current state of the Pregnancy module to determine what the edges should be? Is/was that too expensive?

starsim/demographics.py
```diff
@@ -447,103 +586,237 @@ def init_pre(self, sim):
...
                     raise ValueError(errormsg)
 
                 # Create durations and start dates, and add connections
-                durs = self.dur_postpartum[new_mother_uids]
-                start = np.full(self.n_births, fill_value=ti)
+                durs = self.dur_breastfeed[new_mother_uids]
+                start = np.full(self.n_births_this_step, fill_value=self.ti)
 
                 # # Remove pairs from prenatal network and add to postnatal
                 prenatalnet.end_pairs()
```
RomeshA: It would be nice if performance permits the networks to handle the update by checking who is currently pregnant or breastfeeding, particularly since these blocks are already running for each network independently anyway (in the event that there was more than one). It is a bit confusing that `prenatalnet.end_pairs()` gets called via `update_breastfeeding_network()` for example. If the performance penalty isn't too big to move these to the network `step()` methods I think that would make it a lot simpler and more reliable

starsim/demographics.py
```diff
@@ -447,103 +586,237 @@ def init_pre(self, sim):
...
+        """
+        newborn_parents = self.sim.people.parent[newborn_uids]  # Find which parent each newborn belongs to
+        self.gestation_at_birth[newborn_uids] = self.gestation[newborn_parents]  # Transfer to newborn
+        self.pregnant[uids] = False
+        self.ti_delivery[uids] = self.ti  # Record timestep of delivery as timestep, not fractional time
+        self.gestation[uids] = np.nan  # No longer pregnant so remove gestational clock
+        self.parity[uids] += 1  # Increment parity for the mothers
+
+        # Set maternal death outcomes
+        dead = self.pars.p_maternal_death.rvs(uids)
```
RHull-IDM: `p_maternal_death` is a bernoulli which has `.filter()`. `filter()` returns the uids directly. Then you can skip the `np.any()` check and just use `self.ti_dead[uids[dead]] = self.ti` directly.

starsim/demographics.py
```diff
@@ -447,103 +586,237 @@ def init_pre(self, sim):
...
+        """
+        return
+
+    def set_breastfeeding(self, newborn_uids):
+        """
+        Set breastfeeding durations for new mothers. Thie method could be extended to
+        store duration of exclusive breastfeeding, partial breastfeeding, etc, and these
+        properties could be stored with the infant for tracking other health outcomes.
+        """
+        breastfed_bools = self.pars.p_breastfeed.rvs(newborn_uids)
```
RHull-IDM: can use `p_breastfeed.filter(newborn_uids)` to get `uids` automatically.

RHull-IDM (commented):
I didn't see anything that Romesh missed except for some places where you can use `.filter()` for bernoulli dists to save a few lines of code.

starsim/demographics.py
robynstuart: I thought this might be useful to store explicitly for modules that want to sync the delivery date to seasons, e.g. for seasonal transmission. But I'll remove and users can do their own thing

starsim/demographics.py
robynstuart: Could alternatively argue that we shouldn't call distributions `p_conceive` but rather `dist_conceive` or something. Intuitively I think of `p_conceive` as a scalar between 0 and 1, i.e. it's the parameter that gets fed into the distribution rather than the distribution itself.

starsim/demographics.py
```diff
@@ -306,123 +347,221 @@ class Pregnancy(Demographics):
...
         fertility_rate (float/dataframe): value or dataframe with age-specific fertility rates
         rel_fertility (float): constant used to scale all fertility rates
-        p_maternal_death (float): probability of maternal death during pregnancy (default 0.0)
-        p_neonatal_death (float): probability of neonatal death (default 0.0)
-        sex_ratio (float): probability of female births (default 0.5)
+        p_infertile (bernoulli): probability of primary infertility (default 0)
         min_age (float): minimum age for pregnancy (default 15)
         max_age (float): maximum age for pregnancy (default 50)
         rate_units (float): units for fertility rates (default assumes per 1000)
+        dur_pregnancy (float/dur): duration of pregnancy (drawn from choice distribution by default)
```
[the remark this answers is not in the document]
robynstuart: I've been separating the `Pars` out in lots of places, I feel like it makes it easier to see at a glance what the parameters are. But also, it means that `FPsim` can inherit the `Pars` and only update the ones it needs.

starsim/demographics.py
```diff
@@ -554,42 +827,44 @@ def make_pregnancies(self):
...
-            people.parent[new_uids] = conceive_uids
-            self.child_uid[conceive_uids] = new_uids
-
-            # Add connections to any prenatal transmission layers
-            for lkey, layer in self.sim.networks.items():
-                if layer.prenatal:
-                    durs = np.full(n_unborn, fill_value=self.pars.dur_pregnancy/layer.t.dt) # Network edge duration is stored in units of timesteps, so we need to convert dur_pregnancy to a number of timesteps here
-                    start = np.full(n_unborn, fill_value=self.ti)
-                    layer.add_pairs(conceive_uids, new_uids, dur=durs, start=start)
+            new_uids, new_slots = self._make_newborn_uids(conceive_uids)
```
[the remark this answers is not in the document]
robynstuart: Ok have refactored this, let me know what you think. Now `_make_newborn_uids` takes in two arrays, `conceive_uids` and `embryo_counts`, and returns `conceive_uids_with_repeats, new_uids, new_slots`.

starsim/demographics.py
robynstuart: I've tried refactoring this but I found it messier and slower. It means that the networks have to recalculate a whole lot of things that we just already have ready to go at this point.

Separately there's also the question of where a user would expect something like `dur_breastfeeding` to live. I feel strongly that it should live in the pregnancy module because it can also modulate other things that the pregnancy modules cares about (most importantly rel_sus), and not as an input parameter to the breastfeeding transmission network.

I'll think on it and might try again to shift things, but I think this could also be moved to a subsequent release.

starsim/demographics.py
```diff
@@ -599,20 +874,82 @@ def set_prognoses(self, uids):
...
+        return
 
-        if np.any(dead): # NB: 100x faster than np.sum(), 10x faster than np.count_nonzero()
-            self.ti_dead[uids[dead]] = ti + dur_preg
+    def step(self):
+        if self.ti == 0 and self.pars.burnin:  # TODO: refactor
+            dtis = np.arange(np.ceil(-1 * self.pars.dur_pregnancy.rvs()), 0, 1).astype(int)
+            for dti in dtis:
+                self.t.ti = dti
+                self.do_step()
```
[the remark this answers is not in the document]
robynstuart: I've moved to init_post, I think this is ok to leave as it hasn't caused issues, but this way it's more transparent

Issue #1104, opened 2025-12-12
Bool & UID operators?


RomeshA:
In several places, we treat `BoolArr` as equivalent to `BoolArr.uids` i.e., functions that expect UIDs will treat a BoolArr as representing the true UIDs. It could similarly make sense to have `BoolArr & uids` be equivalent to `BoolArr.uids & uids` (and related operators) which would facilitate logical operations that mix logical indexing on `auids` with collections of `uids`. For example, this can happen when combining states with network edges, like if wanting to find all female participants in a network, users might be expect to be able to write

```
sim.people.female & sim.networks[0].members
```

but this currently would fail because `sim.people.female` is a `BoolState` and `sim.networks[0].members` is a `uids` instance. 

Things to consider in this proposed enhancement would be

- Should the return value be a `BoolArr`, or a `uids` - or could it be either depending on the order of the arguments?
- Is the performance penalty too high for adding a type check to the BoolArr operators?
- Is there scope for this to produce unexpected behaviour in other circumstances?

PR #1108, opened 2026-01-10
Add BoolArr & uids operators for set operations (#1104)

RHull-IDM:
### Description
- Implement bitwise operators (&, |, ^) on `BoolArr` to support operations with `uids` objects. Enables syntax like `sim.people.female & network.members`. Any chaining of `BoolArr` and `uids` will now return uids.

- `uids & BoolArr` was previously supported, so this PR adds support for the reverse order, `BoolArr & uids`.  Also adds explicit errors when attempting to use in-place operations.

- Minimal performance impact: single isinstance() check before fallback

Tests added to tests/test_other.py::test_arrs() verify:
- Set intersection, union, and symmetric difference operations
- Symmetry of operations
- Type consistency (returns uids when operating on uids)
- Preservation of existing BoolArr & BoolArr behavior
- In-place operation rejection

### Checklist
- [X] New tests were needed and have been added

### DIFF
--- a/starsim/arrays.py
+++ b/starsim/arrays.py
@@ -298,13 +298,15 @@ class Arr(BaseArr):
         return
 
     def _boolmath(self, op, other=None, inplace=False):
-        """
-        Helper function for boolean operations
+        """ Helper function for boolean operations.
 
-        :param op: A string with the operator e.g., ">", "&" etc.,
-        :param other: The object operating on the BoolArr e.g., Arr, np.ndarray
-        :param inplace: If True, overwrite the current object rather than returning a new one
-        :return: A BoolArr with the result of the operation
+        Args:
+            op: A string with the operator e.g., ">", "&" etc.
+            other: The object operating on the BoolArr e.g., Arr, np.ndarray.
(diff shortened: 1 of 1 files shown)


RomeshA:
@RHull-IDM @cliffckerr I've updated this PR with a modification to the functionality. I was thinking further that a couple of other issues with the inplace operators were

- It would be useful to have in-place operations with a `BoolArr` and `uids` which would be possible if they returned `BoolArr` instances rather than `uids`, but because the types were changing, it had to explicitly raise an error
- The inplace `BoolArr` operators when operating on other `BoolArr` instances were changing the `id` of the object (I think the `BoolArr` itself though, not the `.raw` array). In general this could become unsafe if a user is doing in-place operations on a `BoolState` as this could break references to the state, if they had been tracked in other modules. 

I've updated this PR so that the operators return `BoolArr` instances if the `BoolArr` is first, e.g.,

```
a = sim.people.female & network.members # Returns a BoolArr
b = network.members & sim.people.female # Returns a uids
```

The in-place operators are also now fully implemented e.g., `sim.people.female &= network.members` is permitted

PR #1111, opened 2026-01-16
Add automatic BoolState results for People (#1098)

### LINKED ISSUE #1098: People boolstates and results
An issue came up where adding a `BoolState` to people directly does not automatically trigger creation of a result, as it would if the state was added to a module. That problem is compounded because although `People` updates results like `n_alive` and `new_deaths` in `People.update_results()`, the creation of those results takes place in `Sim.init_results()` and there is no `People.init_results()` that users can extend.

So this issue would be a two-part fix to

1. Introduce `People.init_results()` so that both creation and updating of the people-associated results is handled by `People` rather than being split across `People` and `Sim`
2. Update `People.init_results()` to have logic similar to `Module.init_results()` such that `BoolState` instances added to `People` automatically get results

RHull-IDM:
### Description

This commit implements automatic result creation for BoolStates added directly to People, matching the behavior already present in Modules.

Changes:
- Add People.auto_state_list property to return BoolStates belonging directly to People (excludes module states)
- Add People.init_results() method to automatically create results for all BoolStates and standard people-related results (new_deaths, etc.)
- Update People.update_results() to automatically update all BoolState results during each timestep
- Refactor Sim.init_results() to delegate result initialization to People.init_results()

This resolves the inconsistency where BoolStates in Modules automatically got results, but BoolStates added directly to People did not.

Fixes #1098

### Checklist
- [X] A new version number was needed & changelog has been updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,12 +4,13 @@ All notable changes to the codebase are documented in this file. Changes that ma
 
 ## Version 3.1.0 (2026-XX-XX)
 - (Planned changes to how pregnancy/births tracks UIDs of mothers and children)
-- Previously `People.(module_name).(state_name)` could be used to access module states (e.g., `people.sir.infected`). However, this mechanism fails to pass on property attributes that are dynamically computed (e.g., `SIR.infectious`). Now `People.(module_name)` simply contains a reference to the module, so all attributes can be accessed. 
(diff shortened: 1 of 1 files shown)


RomeshA:
Thanks @RHull-IDM! Checking the state name for '.' to be able to tell whether it's a 'People' state or a 'Module' state feels a bit fragile to me in the long term, but there doesn't seem to be a better way to do it at the moment! I'll open a separate issue for us to think about that more later

PR #1155, opened 2026-02-16
Add diagnostics

cliffckerr:
### Description
Adds simple diagnostics to store random numbers and/or states.

### Checklist

### DIFF
--- a/starsim/debugtools.py
+++ b/starsim/debugtools.py
@@ -453,7 +453,7 @@ def mock_sim(n_agents=100, **kwargs):
         results = sc.objdict(),
         networks = sc.objdict(), # Needed for infections
         analyzers = sc.objdict(), # Needed for infection log
-        has_diagnostics = False,
+        diagnostics = None,
     )
     return sim
 
--- a/tests/test_debugtools.py
+++ b/tests/test_debugtools.py
@@ -34,7 +34,8 @@ def test_profile():
 @sc.timer()
 def test_diagnostics_rvs():
     sc.heading('Testing sim diagnostics for random numbers')
-    sim = ss.Sim(**debug_pars, diagnostics='rvs')
+    sim = ss.Sim(**debug_pars)
+    sim.set_diagnostics(states=False)
     sim.run()
 
     rvs = sim.diagnostics.rvs


starsim/debugtools.py
```diff
@@ -281,6 +281,83 @@ def run(self, reseed=True):
...
+    @staticmethod
+    def compute_stats(v):
+        """ Compute length, mean, min, and max """
+        stats = sc.objdict(
+            n = len(v),
+            mean = v.mean(),
+            min = v.min(),
+            max = v.max(),
+            hash = sc.sha(v).hexdigest()[:16],
+        )
```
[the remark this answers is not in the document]
cliffckerr: Suggestion is incorrect since all states should be numeric

starsim/sim.py
```diff
@@ -303,6 +305,53 @@ def init_data(self, data=None):
...
+        to file. This allows e.g. understanding exactly where two runs that should
+        have been identical diverged. Results can be exported (in JSON format) and
+        analyzed externally.
+
+        Although the states export could be used for analysis, users are encouraged
+        to write an `ss.Analyzer` instead, as enabling diagnostics significantly degrades
+        sim run time and memory usage.
+
+        Args:
+            which (str): shortcut for setting rvs and/or states; valid args are 'rvs', 'states', and 'both' (equivalent to True)
```
RomeshA: The docstring types are a bit inconsistent, here it's `str` I think elsewhere it has `str/list` (in `sim.py`) and `bool` is actually also supported in the code?

RomeshA: I think it might be cleaner if the `Diagnostics.__init__()` arguments more directly aligned with `sim.set_diagnostics()` i.e., if it was `Diagnostics.__init__()` that handled the `which` argument.

But actually, related to my other comment about directly calling `sim.diagnostics.export()`, if that was the point at which the file names were set, then the `rvs` and `states` arguments could also be dropped entirely, if the user exclusively sets them via `which`

starsim/debugtools.py
```diff
@@ -281,6 +281,78 @@ def run(self, reseed=True):
         return
 
 
+class Diagnostics(sc.quickobj):
+    """ Helper class for storing detailed diagnostic information; see sim.set_diagnostics() for usage; not to be used directly
+
+    See also ss.Debugger() for a user API for debugging multiple sims.
+    """
+    def __init__(self, sim, rvs=False, states=False, export=None, detailed=False):
```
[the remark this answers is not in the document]
cliffckerr: Done

starsim/debugtools.py
```diff
@@ -281,6 +281,78 @@ def run(self, reseed=True):
...
+        self.sim = sim
+        self.do_export = export or isinstance(rvs, str) or isinstance(states, str)
```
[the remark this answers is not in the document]
cliffckerr: Done

starsim/sim.py
cliffckerr: Args removed -- just `rvs` and `states` now

cliffckerr: Agree!

PR #1166, opened 2026-02-20
Fix ASFR annualization being overwritten by zeros

robynstuart:
Remove duplicate `asfr = np.zeros(...)` and redundant `tdim` recomputation in Pregnancy.finalize() that overwrote the correctly computed rolling annual sum with an array of zeros.

### Checklist
- [ ] All new functions have a docstring and are appropriately commented
- [ ] New tests were needed and have been added, or no tests required
- [ ] Changelog has been updated, or there are no user-facing changes


cliffckerr (commented):
I'm so sorry, this was a mangled merge on my part! Thanks for fixing!

PR #1182, opened 2026-02-27
Get module

### LINKED ISSUE #1175: Write ndict.get() and sim.get_module()
```py
sim.get_module('tb') # works by name
sim.get_module(tbsim.TB) # works by type
sim.diseases.get(tbsim.TB) # also works at the ndict level
```
These would return the first match, could have maybe `sim.get_modules()` that would return all matches? Or should it return an error if more than one match?

RomeshA:
### Description

Adds `sim.get_module` and `sim.get_modules()`. If the `Sim` is not initialized, modules will be retrieved from `sim.pars` instead.

This only partially closes #1175 as `ndict.get` has not been implemented in this PR. However, I would imagine that in most cases `ndict.get` could probably be replaced by `sim.get_module` for the purpose of retrieving modules (per the example `sim.diseases.get(tbsim.TB)` where `sim.get_module(tbsim.TB)` would probably do the same thing?) so `ndict.get()` might need to be coded against a broader range of use cases. 

### Checklist
- [X] All new functions have a docstring and are appropriately commented
- [X] New tests were needed and have been added, or no tests required
- [X] Changelog has been updated, or there are no user-facing changes

### DIFF
--- a/starsim/networks.py
+++ b/starsim/networks.py
@@ -995,9 +995,9 @@ class BreastfeedingNet(PostnatalNet):
 
     def init_post(self, *args, **kwargs):
         # Connect the pregnancy module to this network
-        for module in self.sim.modules:
-            if isinstance(module, ss.Pregnancy):
-                self.pregnancy = module
+        for mod in self.sim.modules:
+            if isinstance(mod, ss.Pregnancy):
+                self.pregnancy = mod
                 break
         else:
             raise RuntimeError('BreastfeedingNet requires a Pregnancy module in the simulation to track breastfeeding status')
--- a/starsim/sim.py
+++ b/starsim/sim.py
@@ -185,7 +185,7 @@ class Sim(ss.Base):
             self.diagnostics.store_states(key='init')
         return self
 
-    def get_module(self, query):
+    def get_module(self, query, die=True, match_case=False):
         """
         Retrieve a single module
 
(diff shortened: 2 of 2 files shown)


starsim/networks.py
```diff
@@ -995,7 +995,7 @@ def __init__(self, pars=None, **kwargs):
 
     def init_post(self, *args, **kwargs):
         # Connect the pregnancy module to this network
-        for module in self.sim.module_list:
+        for module in self.sim.modules:
```
cliffckerr: Should use `for mod` for consistency

starsim/sim.py
```diff
@@ -187,6 +185,62 @@ def init(self, force=False, timer=False, **kwargs):
             self.diagnostics.store_states(key='init')
         return self
 
+    def get_module(self, query):
+        """
```
cliffckerr: Looks good, but let's add a `die` arg

starsim/modules.py
```diff
@@ -82,6 +86,7 @@ def find_modules(key=None, flat=False, verbose=False):
     # Initialize dict
     for modkey in modmap.keys():
         modules[modkey] = sc.objdict()
+    modules['modules'] = sc.objdict() # Catch-all for modules that ship with Starsim but that do not match a standard type
```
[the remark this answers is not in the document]
cliffckerr: Done

PR #1184, opened 2026-02-27
add action to run tests weekly

### LINKED ISSUE #1181: Update CI testing
- [ ] Add nightly tests -- can just copy this file and edit the paths: https://github.com/sciris/sciris/blob/main/.github/workflows/schedule_tests.yaml
- [ ] Look into installing dev versions / auto-reading changelogs for potentially breaking changes
- [ ] Rename `test_spokes.yaml` to `test_models.yaml`

RHull-IDM:
Add weekly test runs and rename test_spokes to test_models. Closes #1181

### DIFF
--- a/.github/workflows/cron_tests.yaml
+++ b/.github/workflows/cron_tests.yaml
@@ -2,7 +2,7 @@ name: Starsim cron tests
 on:
   workflow_dispatch:
   schedule:
-    - cron: "0 0 * * 0" # Run every Sunday at midnight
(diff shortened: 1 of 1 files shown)


cliffckerr (commented):
Looks good, I think it's fine to run nightly rather than weekly. A week can be a long time :) (I'd forgotten that the Sciris tests only run once a week, but I think daily is better for Starsim since higher churn + touches more things so more likely to break.)

PR #1225, opened 2026-03-09
Calibration updates

RomeshA:
Draft PR for changes to calibration - the key functionality being added is to 

- Provide built-in functionality for running multiple seeds per trial, so users can write a `build_fn` that returns a `Sim`, set `n_reps=10` for example, and this will automatically run 10 seeds per trial and average results over seeds
- Have multiple seeds run in parallel, with a particular focus on being able to request say, 10 seeds per trial and 120 CPUs, and have it run 12 workers with 10 CPUs each
- Facilitate parallelization across multiple machines, to have a straightforward and clear workflow for running a single calibration across several VMs

One other change in this PR is to remove the random seed from being an optimization parameter but rather to have it generated deterministically as the optimization progresses. The intention (although not verified) is to improve Optuna convergence by reducing the dimensionality of the optimization, particularly noting that the random seed is meant to be uncorrelated with the model output anyway. 

The features above are all implemented and are currently being tested in some applications now. Aside from debugging/review of the concept, remaining tasks are

- Updates to documentation and example workflows
- Check what changes are needed for `CalibComponent` as some of the components expect the random seed to have a particular format which may be different now


RomeshA:
Yes would be good to get feedback on that, it still varies the seed and tracks/records the seed in the output, it's just that the seed is being set deterministically based on the trial number and repetitions rather than being sampled. I was thinking that since the outputs are meant to be uncorrelated with regard to the seeds it shouldn't matter whether they're selected sequentially/deterministically or sampled, but would be curious if that causes issues for some of those workflows

PR #1227, opened 2026-03-09
Arr fix

RomeshA:
### Description

An error was encountered in an application where a module contained

```
        self.n_contacts_12 -= contacts
```

where `n_contacts_12` and `contacts` are both `FloatArr` instances. The error was caused by an infinite recursion in `BaseArr.__array_ufunc__` resulting in Python crashing. Once this error was fixed, a different error occurred

```
AttributeError: Cannot modify attribute "n_contacts_12";
```

This occurs if a new array is being written to the attribute, but the in-place operation should not create a new array. Previously this error was unreachable because the infinite recursion would crash Python first. 

The tests didn't pick this up because the in-place functionality was only tested using `BoolArr` which implementes its own versions of the in-place operators (due to handing mixed `BoolArr`/`uids` operations). This PR also adds an additional test to cover `FloatArr`

### Checklist

### DIFF
--- a/starsim/people.py
+++ b/starsim/people.py
@@ -334,6 +334,11 @@ class People:
         return ~self.alive
 
     @property
+    def born(self):
+        """ Born boolean """
+        return self.age >= 0
+
+    @property
     def male(self):
         """ Male boolean  (female is a state) """
         return ~self.female


RomeshA:
Converting this back to draft as it looks like it doesn't work when there are inactive agents because then `.values` and `.raw` don't match. Might require a more extensive solution

PR #1242, opened 2026-03-12
Fix HouseholdNet.add_pairs UID indexing

daniel-klein:
## Summary
- `household_ids[-1]` fails because `FloatArr` is UID-indexed and `-1` is not a valid UID
- Use the `cluster_id` counter instead, which already tracks the last assigned household ID after the while loop

## Context
Found while building the [starsim-community-structure](https://github.com/starsimhub/starsim-community-structure) repo which subclasses `HouseholdNet`.

## Test plan
- Run any sim with `ss.HouseholdNet(dynamic=True)` — previously crashes on init, now works

### DIFF
--- a/starsim/networks.py
+++ b/starsim/networks.py
@@ -1115,6 +1115,7 @@ class HouseholdNet(Network):
             ]
         self.define_states(*states)
         self.p_fractional_age = ss.uniform()
+        self.n_households = 0 
         return
 
     def init_pre(self, sim):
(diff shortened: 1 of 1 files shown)


starsim/networks.py
```diff
@@ -1179,7 +1179,7 @@ def add_pairs(self):
 
         if self.dynamic:
             # Find a female head of household between ages 15 and 50
-            for cid in range(int(self.household_ids[-1])):
+            for cid in range(cluster_id + 1):
```
cliffckerr: I don't love using an incidental loop variable for this. But in addition, it looks like we're doing a slow comparison (`ss.uids(self.household_ids == cid)`) to reconstruct the `cluster_uids` variables that we just calculated above. Is it possible to avoid doing this, even if we have to store this info twice, unless that would require a slow and awkward step to keep the two representations in sync?

PR #1309, opened 2026-04-08
Fix dynamic household births

RHull-IDM:
Rework add_births() so that newborns receive a household ID exactly once when they are delivered (age >= 0 and parent set), rather than using an age-window heuristic tied to update_freq. Move add_births() before the update_freq gate so it runs every timestep, ensuring it works correctly with any update frequency or when the network timestep exceeds gestation.

Also fix self-loops, assign household IDs before creating edges, and update the test to account for undelivered embryos at end-of-sim.

### Checklist
- [ ] All new functions have a docstring and are appropriately commented
- [ ] New tests were needed and have been added, or no tests required
- [X] Changelog has been updated, or there are no user-facing changes


cliffckerr (commented):
Looks good!

PR #1312, opened 2026-04-08
Add apply_age_range and extend parse_age_range with bracket notation

robynstuart:
I don't _think_ this function exists already but pls correct me if it does!

### DIFF
--- a/starsim/utils.py
+++ b/starsim/utils.py
@@ -245,45 +245,120 @@ def find_contacts(p1, p2, inds):  # pragma: no cover
 
 def parse_age_range(age_string) -> tuple:
     """
-    Parse an age range string into lower and upper bounds
+    Parse an age range string into lower and upper bounds.
 
-    Example usage:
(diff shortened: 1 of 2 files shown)


starsim/utils.py
```diff
@@ -245,47 +245,122 @@ def find_contacts(p1, p2, inds):  # pragma: no cover
...
-        - '<5' - which returns (0.0, 5.0)
-        - '95+' - which returns (95.0, np.inf)
-        - '>95' - which returns (95.0, np.inf)
+    Args:
+        age_string: a string specifying an age range
 
-    :param age_string: A string specifying an age range (e.g., '5-9', '5 to 9', '95+')
-    :return: A tuple with (lower, upper) bound values.
+    Returns:
+        A tuple ``(lower, upper)`` as floats.
```
cliffckerr, line 277: With Quarto you only need single backticks, but can fix that later.

PR #1347, opened 2026-06-03
Clear caches

### LINKED ISSUE #1343: Starsim memory "leak"?
Tess was experiencing out-of-memory issues when running large number of TBsim simulations for history matching. Working with Claude, we traced it back to the following:

> Root cause: a reference leak in starsim v3.3.3 — each ss.Sim's start_step/finish_step bound methods are registered into a module-level list and never removed, so every sim a worker builds stays pinned in memory (with its ~41 MB network edge arrays + distributions + state). With sc.parallelize not recycling workers, each of the 120 workers accumulates ~20 sims/wave → ~2–4 GB each → ~290 GB → OOM. The smooth, never-recovering climb in your graph is exactly this within-wave accumulation.

I have not further validated that this is the cause, but it seems plausible considering hope loop.py is configured.

RomeshA:
### Description

Relating to the discussion in #1343 this proposes a more targeted clearing of content. For our memory leak case, there were two places causing problems

- `Dist._callable_args` containing references back to the `Sim` and `uids` etc.
- `sim.loop.plan` containing references to the `Sim` inside a dataframe, preventing garbage collection

Running `shrink(full=False)` is still potentially problematic in cases where the user has run with `sim.run(until=)` and also if an error is raised to terminate early. This PR proposes a fix in two parts

- `Dist._callable_args` is a temporary cache that will be automatically regenerated if it is cleared. Therefore, `sim.run()` can use a `finally` block to guarantee this cache is cleared when execution stops, and the `Sim` can just be directly resumed and the caches will be repopulated as required, no action required by the user
- `sim.loop.plan` is difficult to regenerate because the user may have called `.insert()` arbitrarily after initialization or during execution. However, it would not be necessary to clear the plan if the plan did not live within a dataframe. So this PR proposes one possible refactoring to have the plan be stored in pure Python containers so that garbage collection works without any complications, and then has a method to generate a dataframe representation of the plan on-demand

Together, these appear to resolve the memory leak while also correctly handling early termination, although it would be interesting if there are other scenarios where the leak persists. If proceeding with this, we could possibly remove the `full` argument from `shrink` so it would be clearer what the role of `shrink` is

### DIFF
--- a/starsim/distributions.py
+++ b/starsim/distributions.py
@@ -163,6 +163,13 @@ class Dists(sc.prettyobj):
             out += dist.reset()
         return out
 
+    def clear_run_caches(self):
+        """ Clear rebuildable per-run caches from each managed distribution """
+        if self.dists is not None:
+            for dist in self.dists.values():
+                dist.clear_run_cache()
+        return
+
     def copy_to_module(self, module):
         """ Copy the Sim's Dists object to the specified module """
         matches = {key:dist for key,dist in self.dists.items() if id(dist.module) == id(module)} # Find which dists belong to this module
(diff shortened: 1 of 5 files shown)


cliffckerr (approved):
Looks good, thanks, sorry for the delay!

PR #1394, opened 2026-07-11
Multi random dtype

### LINKED ISSUE #1395: Multi-random return argument length
`multi_random` returns an incorrect length in 3.5.2 depending on the dtype - but out of the box, in 3.3.4:

```python
>>> import numpy as np, starsim as ss
... sim = ss.Sim(n_agents=50, diseases='sir', networks='random', dur=ss.years(1)).init()
... mr = ss.multi_random('source', 'target')
... mr.init(sim=sim)
... p1 = ss.uids(np.arange(10))
... p2 = ss.uids(np.arange(10, 20))
... print(len(mr.rvs(p1, p2)))# -> 20(BUG: should be 10)
... print(mr.dists[0].rvs(p1).dtype)# -> float64 (but `random` declares float32)
...
Initializing sim with 50 agents
10
float32
```

but in 3.5.1:

```python
>>> import numpy as np, starsim as ss
... sim = ss.Sim(n_agents=50, diseases='sir', networks='random', dur=ss.years(1)).init()
... mr = ss.multi_random('source', 'target')
... mr.init(sim=sim)
... p1 = ss.uids(np.arange(10))
... p2 = ss.uids(np.arange(10, 20))
... print(len(mr.rvs(p1, p2)))# -> 20(BUG: should be 10)
... print(mr.dists[0].rvs(p1).dtype)# -> float64 (but `random` declares float32)
...
Initializing sim with 50 agents
20
float64
```

### LINKED ISSUE #1396: ss.random() dtype changes depending on UIDs
The return dtype for `ss.random()` is different depending on whether it is called with or without UIDs. 

```
sim = ss.Sim(n_agents=50, diseases=ss.SIR(dur_inf=ss.random())).init()
Initializing sim with 50 agents
sim.diseases[0].pars.dur_inf.rvs().dtype
Out[22]: dtype('float32')
sim.diseases[0].pars.dur_inf.rvs([1]).dtype
Out[23]: dtype('float64')
```

As `ss.random()` has a `dtype` parameter, it should return the requested dtype for all `rvs` calls

### LINKED ISSUE #1397: ss.poisson() dtype changes depending on UIDs
Similar to #1396 , `ss.poisson` also returns variable dtypes

```
sim = ss.Sim(n_agents=50, diseases=ss.SIR(dur_inf=ss.poisson())).init()
Initializing sim with 50 agents
sim.diseases[0].pars.dur_inf.rvs().dtype
Out[6]: dtype('int64')
sim.diseases[0].pars.dur_inf.rvs([1]).dtype
Out[7]: dtype('float64')
```

### LINKED ISSUE #1398: Minimum python should be 3.11 due to `networkx`
Currently `uv` complains about `pyproject.toml` specifying that Starsim supports Python >3.10, when we also declare a dependency on `networkx` which only supports Python >3.11. So logically we must also require at least Python 3.11. Recording this as a bug because `uv` workflows report errors due to this mismatch. But also just to note that Python 3.10 is EOL in October anyway, so I would propose we just update our minimum Python version to 3.11 to match

### LINKED ISSUE #1400: `multi_random` incorrect dists
`multi_random` doesn't create the intended number/names of distributions e.g., 

```
[x.name for x in ss.multi_random(["a", "b"], "c", "d").dists]
['a', 'b', ('c', 'd')]
```

This is because it contains

```
    def __init__(self, names, *args, crn=None, **kwargs):
        names = sc.mergelists(names, args)
```

but `args` is a `tuple` not a `list` and therefore doesn't get merged properly. Easiest fix is probably to change it to `*args` to unpack the args

RomeshA:
### Description

`multi_random` returns an incorrect length in 3.5.1 depending on the dtype - but out of the box, in 3.3.4:

```python
>>> import numpy as np, starsim as ss
... sim = ss.Sim(n_agents=50, diseases='sir', networks='random', dur=ss.years(1)).init()
... mr = ss.multi_random('source', 'target')
... mr.init(sim=sim)
... p1 = ss.uids(np.arange(10))
... p2 = ss.uids(np.arange(10, 20))
... print(len(mr.rvs(p1, p2)))# -> 20(BUG: should be 10)
... print(mr.dists[0].rvs(p1).dtype)# -> float64 (but `random` declares float32)
...
Initializing sim with 50 agents
10
float32
```

but in 3.5.1:

```python
>>> import numpy as np, starsim as ss
... sim = ss.Sim(n_agents=50, diseases='sir', networks='random', dur=ss.years(1)).init()
... mr = ss.multi_random('source', 'target')
... mr.init(sim=sim)
... p1 = ss.uids(np.arange(10))
... p2 = ss.uids(np.arange(10, 20))
... print(len(mr.rvs(p1, p2)))# -> 20(BUG: should be 10)
... print(mr.dists[0].rvs(p1).dtype)# -> float64 (but `random` declares float32)
...
Initializing sim with 50 agents
20
float64
```

The issue is twofold

1. `ss.random.rvs()` returns a `float64` even if its dtype is set to `float32` because the casting to the target dtype is skipped if `rvs()` is called with UIDs. For example

    ```sim = ss.Sim(n_agents=50, diseases=ss.SIR(dur_inf=ss.random())).init()
    Initializing sim with 50 agents
    sim.diseases[0].pars.dur_inf.rvs().dtype
    Out[22]: dtype('float32')
    sim.diseases[0].pars.dur_inf.rvs([1]).dtype
    Out[23]: dtype('float64')
    ```
    
2. `multi_random` casts `float` to `int` in order to apply the XOR step. However, implicitly it assumes that the float and int widths match. By default this is _supposed_ to be true (since the default `float` is `np.float32` and the default `rand_uint` is `np.uint32`). The bug here has been caused by `ss.random.rvs()` returning a `float64` which doesn't match. But a future bug could occur if the user changes `ss.dtypes.float` without making a corresponding change to the integer types

This PR implements two fixes

1. Ensure `ss.random.rvs()` returns the specified dtype in all cases
2. The `float-int-float` round-trip in `multi_random` is an internal step for performing the `XOR` combination. Therefore, the intermediate integer type should be selected to match the `float` dtype rather than using `ss.dtypes.rand_uint`. This removes the need for the user to ensure the types match

However, the fix is not as straightforward as just casting the float64 to float32, because it is supposed to generate numbers on the open interval `[0,1)` but just downcasting can result in the `float32` taking a value of 1, which can be problematic for dists like `expon`. Instead, the target dtype needs to be identified and used by `hash_uniform` directly.

In fixing the return dtype for `ss.random` it was also identified that `ss.poisson` returned different dtypes depending on whether UIDs were provided. It has similarly been updated to always return `ss.dtypes.int`.

### Checklist


cliffckerr (approved):
Looks great, thanks for catching all these!

